PR #34, opened 2015-01-07
Add prospector-mirror

guykisel:
In landscapeio/prospector#78, Prospector was updated to allow multiple command line file inputs, which makes it easy to create a pre-commit hook for Prospector. The hook is configured to run as a system hook rather than a python hook to allow pylint to validate imports.

https://github.com/guykisel/prospector-mirror


asottile:
Looks like your hooks.yaml has some invalid syntax.  Probably needs quotes around a string literal.

fixnship :)

guykisel:
Whoops, I added a description and mistakenly assumed it was a harmless change.

guykisel:
Fixed: https://github.com/guykisel/prospector-mirror/commit/00fbd80101566b1b9c873c71f2ab7b95b8bd0a7d

PR #38, opened 2015-03-07
Add https://github.com/hootsuite/pre-commit-php

asottile:
@shihoonhootsuite thanks for the great article: http://code.hootsuite.com/using-yelps-pre-commit-hooks/


struys:
@shihoonhootsuite We'd love to get PHP officially supported without needing system level hooks. Any suggestions/pull-requests would be great! - https://github.com/pre-commit/pre-commit/issues/209

PR #39, opened 2015-03-07
Add puppet-lint

asottile:


struys:
This is working on osx but i'm having troubles getting it working in linux. It might just be specific to lucid - https://github.com/rodjek/puppet-lint/issues/364

PR #40, opened 2015-03-12
Update index.mako

struys:
@guykisel has been contributing a lot he should totally be on this list! Anyone else we should add?


guykisel:
:)

PR #42, opened 2015-04-08
Adds JSCS to repos.

hyperNURb:


asottile:
We have a tool that helps create these repos if you'd like to use that:  https://github.com/pre-commit/pre-commit-mirror-maker

hyperNURb:
Sweet, thanks!

PR #57, opened 2015-07-28
Add chriskuehl/puppet-pre-commit-hooks

chriskuehl:
https://github.com/chriskuehl/puppet-pre-commit-hooks

3 hooks:

* puppet-validate: uses `puppet parser validate` to validate manifests
* erb-validate: compiles erb templates and checks them with `ruby -c`
* puppet-lint

The last obviously overlaps with the puppet-lint mirror, but it seems reasonable to keep these three in one repo.


asottile:
Thanks again!

PR #63, opened 2015-09-24
Adding golang hooks

dnephin:
I looked at the SamWhited/go-pre-commit.git hooks, but they don't quite work for larger repos (`go vet` fails if you pass it files from multiple packages, the `go fmt` hook hides all output) so I created these.


asottile:
Thanks for the PR, friend!

PR #69, opened 2015-11-23
Added the `additional_dependencies` attribute to docs

blarghmatey:
@asottile I'm happy to modify the wording as necessary.


asottile:
Thanks again @blarghmatey!

PR #72, opened 2016-01-06
Fix responsive layout by allowing <code> to wrap

asottile:
@mfnkl @theresama

Before:
Before

After:
After


theresama:
:shipit: cat

PR #76, opened 2016-05-24
add jstewmon/check-swagger

jstewmon:


asottile:
Looks great!  Thanks for adding a hook (I'll merge once travis is happy)

Just a note, `exclude` is supported natively through pre-commit's configuration: (^F exclude: http://pre-commit.com/)

PR #82, opened 2016-09-28
Added antonbabenko/pre-commit-terraform

antonbabenko:
pre-commit hook which formats Terraform configuration files


asottile:
Thanks!

PR #89, opened 2017-01-06
Add git@github.com:troian/pre-commit-golang.git

troian:
Signed-off-by: Artur Troian <troian.ap@gmail.com>


asottile:
Looks good!  One cool thing that we added in the latest release that would reduce (or eliminate!) the external dependencies with these hooks is pre-commit docker support:
- https://github.com/pre-commit/pre-commit/blob/cc0f40fc9666c9f9d06498c489d8f7e5867a5261/CHANGELOG.md#0100 
- http://pre-commit.com/#supported-languages

One slight thing, your readme suggests `sha: HEAD` which _can_ cause issues: https://github.com/pre-commit/pre-commit/issues/158#issuecomment-54103765 (it'll work, but it won't point to the latest if master progresses)

PR #90, opened 2017-01-11
Add pre-commit-clang-format

fdhadzh:
Add clang-format hook for pre-commit


asottile:
lg2m

One thing we added in a recent version was docker support: https://github.com/pre-commit/pre-commit/blob/cc0f40fc9666c9f9d06498c489d8f7e5867a5261/CHANGELOG.md#0100

This hook could also be implemented using that so users wouldn't need a system dependency

asottile:
Thanks!

PR #93, opened 2017-01-24
Change package repo

fdhadzh:


asottile:
Please branch off of `real_master` and target your merge against that.

asottile:
I went ahead and did this for you as it was blocking some work I was doing :)

PR #98, opened 2017-03-10
Add pydocstyle hook

jpulec:
Seems that github.com/FalconSocial/pre-commit-mirrors-pep257 hasn't been updated in a year and for `pydocstyle` so this repo exists.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -41,3 +41,4 @@
 - git://github.com/antonbabenko/pre-commit-terraform
 - git://github.com/willthames/ansible-lint
 - git://github.com/doublify/pre-commit-clang-format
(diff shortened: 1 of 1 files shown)


asottile:
Thanks!

Another note (that we haven't quite decided on whether to deprecate mirror repositories yet or not) is that mirror repositories are less necessary after 0.13.0 [repository local hooks](http://pre-commit.com/#repository-local-hooks) support installing libraries directly.

So instead of a mirror repository you can do something like this:

```yaml
- repo: local
  hooks:
  -   id: pydocstyle
      name: pydocstyle
      entry: pydocstyle
      files: \.py$
      language: python
      additional_dependencies: ['pydocstyle==...']
      # etc.
```

PR #101, opened 2017-03-23
Add pre-commit-go

fdhadzh:


asottile:
Can you sort this next to the other doublify repository?

Also there seem to be two or three other repositories which also provide `go vet` and `go fmt` -- are those ones different / unusable?

Thanks for the PR :D

fdhadzh:
Yes, we know about similar packages, but this package is under our support.

PR #111, opened 2017-05-13
Add pass_filenames hook option

absassi:
This PR updates documentation according to the changes in the corresponding [PR 538 in pre-commit/pre-commit](https://github.com/pre-commit/pre-commit/pull/538).

I took the liberty to exclude the "An executable must satisfy the following things:" phrase and the list formatting since now there is only one requirement for the hook. I've also assumed the version incorporating this change will be 0.14.0, but in case I'm wrong, then the table must be updated.


asottile:
Thanks!

PR #113, opened 2017-06-14
Fix typo

chriskuehl:


asottile:
yay the build is broken >.<

asottile:
seems to be something that changed in https://github.com/dahlia/libsass-python/releases/tag/0.13.0 so I guess that's my problem to fix!

asottile:
Here's the upstream issue: https://github.com/dahlia/libsass-python/issues/208

This PR looks good though 👍

PR #132, opened 2017-09-17
Write more markdown, less html

asottile:
I for one keep making html mistakes!

This swaps bits of the html out for inline markdown, which imo is both easier to _write_ and _edit_.

I only did some parts of the file -- if this seems good I'll do more of the file and add a few features to the markdown parser such that I can eliminate even more html blocks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
 *.html
-*.iml
+*.pyc
 *.swp
-.DS_Store
-.idea
-.project
-.pydevproject
 /all-hooks.json
 /node_modules
 /build
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 aspy.yaml
 mako
 libsass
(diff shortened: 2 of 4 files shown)


asottile:
CC @Lucas-C @chriskuehl

chriskuehl:
html and markdown are both such legacy technologies. personally i think it'd be less error-prone if you re-wrote this using react components.

.gitignore
```diff
@@ -1,10 +1,6 @@
 *.html
-*.iml
+*.pyc
 *.swp
-.DS_Store
-.idea
-.project
-.pydevproject
```
chriskuehl: +++

requirements-dev.txt
```diff
@@ -1,5 +1,6 @@
 aspy.yaml
 mako
 libsass
+markdown-code-blocks
```
chriskuehl: nice

PR #134, opened 2017-09-18
Document install-local does not function on windows

### LINKED ISSUE #131: Document install-local.py is not supported on windows
- [ ] produce an error message in install-local.py if it is unlikely to work as expected (missing `os.symlink`, paths in virtualenvs are not `bin` but instead `Scripts`
- [x] Call out that `pip` should be used on windows to install pre-commit in the documentation

Originally from https://github.com/pre-commit/pre-commit/issues/620

asottile:
Resolves #131

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -62,7 +62,9 @@ pip install pre-commit
 
 Non Administrative Installation:
 
-_(To upgrade: run again, to uninstall: pass `uninstall` to python)_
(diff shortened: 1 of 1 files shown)


asottile:
Actually, `hasattr(os, 'symlink')` is insufficient due to changes in python3, I think I'll just drop the script change and only mention it in the documentation.

asottile:
CC @Lucas-C

PR #145, opened 2017-11-01
Add beautysh to the list of repos

jesselang:
A Bash beautifier for the masses. https://pypi.python.org/pypi/beautysh


asottile:
This is awesome, thanks!

PR #149, opened 2017-12-15
Remove pricematch mirrors and hooks

tdeo:
Hi,

We've received a PR on one of those mirrors https://github.com/pricematch/mirrors-rubocop/pull/3, we're actually not maintaining them anymore, and they've become pretty obsolete with support for local hooks in other languages.


asottile:
Thanks again :)

PR #155, opened 2018-02-01
Document the fact that the hook's name can be overriden

bagerard:
Just realized that it was possible to override the name of a hook. I think it's an interesting feature, so I thought I'd document it.


asottile:
yep, definitely a cool side-effect of the way the configuration gets merged.  I've mentioned this a few times in issue comments but it's nice to get this called out in the docs!

Thanks for the PR :)

PR #161, opened 2018-04-26
Document log_file hook option

asottile:


asottile:
simply forgot to document this!

PR #163, opened 2018-05-21
Add new hooks for Ansible and Docker checks

IamTheFij:
Add two new hook repos. One to contain some Ansible check features and the other for Docker.

**Ansible**
Currently contains a check that vault files are encrypted before commit.

**Docker**
Contains a check to validate `docker-compose.y[a]ml` files are valid by using `docker-compose config`


asottile:
Thanks!

PR #166, opened 2018-05-27
Fix typo under python_venv

chriskuehl:


asottile:
lol woops -- thanks for the fix :D

PR #171, opened 2018-06-19
Add mattlqx repos

mattlqx:
Add some repos I created for our local requirements that might be useful for others.


asottile:
Awesome, thanks!

PR #174, opened 2018-07-18
add search-and-replace plugin

mattlqx:
Plugin to search and replace arbitrary strings in the commits. Handy for when you have a mirror of a repo you'd rather be using, or enforcing copyright dates or whatever!


asottile:
Neat! I was actually thinking of implementing something similar to this directly in the framework at some point (similar to how the [pygrep](https://pre-commit.com/#pygrep) language works today)

PR #175, opened 2018-07-19
Added bashate hook

ssbarnea:


ssbarnea:
@asottile Now the bashate added the hook definiton file and made a new release we should be able to document its existence. :)

PR #177, opened 2018-07-30
Add pre-commit/commitlint hook

alessandrojcm:


all-repos.yaml
```diff
@@ -77,3 +77,4 @@
 - https://github.com/mattlqx/pre-commit-sign
 - https://github.com/mattlqx/pre-commit-search-and-replace
 - https://github.com/openstack-dev/bashate
+- https://github.com/alessandrojcm/commitlint-pre-commit-hook
```
asottile: can't help but notice the irony that you didn't run `pre-commit` while committing this 😆

asottile:
I wonder if we should pursue adding metadata directly to that repository -- or if you should look at https://github.com/pre-commit/pre-commit-mirror-maker

alessandrojcm:
Didn't see that one, I'll check it out.

PR #180, opened 2018-08-29
Added speshak/pre-commit-cfn-lint

speshak:


asottile:
There's discussion (and reasons why `cfn-lint` doesn't currently work with the framework) in the upstream issue here: https://github.com/awslabs/cfn-python-lint/pull/167 -- are these addressed by your repo?

speshak:
Nope, I missed that WIP when I banged this out.  I'll close this in favor of waiting for upstream to fix this correctly.

asottile:
:tada:  thanks again for the PR though!

speshak:
In the off chance someone comes across this in the future, I updated my pre-commit-cfn-lint project to work around the limitation of `cfn-lint`.  I still think that waiting for the upstream project to correct the problem is the right answer, my wrapper is just a cheap workaround until that is done.

PR #193, opened 2018-12-21
Drop system level installation instructions

### LINKED ISSUE #190: Don't suggest running get-pip.py with sudo
The regular caveats of running pip with sudo apply, including (but not limited to) breaking the user's installation of pip.

https://stackoverflow.com/a/37104948/1931274

pradyunsg:
Fixes #190


asottile:
I think we can just delete the whole "System install" section probably (delete lines 72-77)

PR #196, opened 2018-12-27
Update docs to include the alias parameter.

s0undt3ch:


index.mako
```diff
@@ -164,6 +164,14 @@ repository's configuration.
 =r=
     =c= `id`
     =c= which hook from the repository to use.
+=r=
+    =c= `alias`
+    =c= (optional) Provide an alias to the hook. The `alias` allows
+        differentiation for when one wants to include the same hook more
+        than once, with, for example, different parameters. This allows
+        `pre-commit run <alias>` to target a specific entry of the hook to
+        run.
```
asottile: I'd say something like

(optional) allows the hook to be referenced using an additional id when using `pre-commit run <hookid>`.

(this matches the capitalization and is brief and too the point)

s0undt3ch: Updated

asottile:
Thanks :tada:

s0undt3ch:
Anytime!

PR #200, opened 2019-01-07
Swap mirrors for official PyCQA repos

1oglop1:
Mirrors are probably no longer necessary since it's not guaranteed that they will stay up to date.

This was the case especially with `pydocstyle`


1oglop1:
I guess that would be nice, if it's not a lot of work.

asottile:
There's only 2 to do, could you handle it?

PR #204, opened 2019-01-10
Document installation from conda-forge

asottile:


asottile:
@nicoddemus does this look right?

index.mako
```diff
@@ -79,9 +79,17 @@ pre-commit
...
-```
+```bash
 brew install pre-commit
 ```
+
+Using [conda](https://conda.io) (via [conda-forge](https://conda-forge.org)):
+
+```bash
+conda config --add channels conda-forge
+conda install pre_commit
```
nicoddemus: I think it is better to advertise:

```
conda install -c conda-forge pre_commit
```

This is what's suggested [here](https://anaconda.org/conda-forge/pre_commit), also the former would change the global configuration which is not very nice.

asottile: nice! pretty clear that

asottile:
@nicoddemus thanks for the review!

nicoddemus:
sure, anytime!

asottile:
@nicoddemus btw the directions I grabbed were from https://github.com/conda-forge/pre_commit-feedstock#installing-pre_commit

nicoddemus:
Hmm I see. That's probably because conda-forge wants to advertise for users to switch to its channel for good. 👍

PR #205, opened 2019-01-11
Document `require_serial`

asottile:


index.mako
```diff
@@ -367,6 +367,10 @@ file that tells pre-commit:
     =c= `pass_filenames`
     =c= (optional: default `true`) if `true` this hook must take filenames as
         positional arguments. _new in 0.14.0_.
+=r=
+    =c= `require_serial`
+    =c= (optional: default `false`) if `true` this hook will execute using a
+        a single process instead of in parallel. _new in 1.13.0_.
```
cancan101: what exactly does running in parallel mean? Does that mean at the same time as other hooks? How does it work if some of the hooks are marked as require_serial true and others false ?

asottile: They never cross over other hooks, that would almost never work with fixers.

PR #207, opened 2019-01-31
Added pip-compile-multi

nschlemm:
Caters for verifying that pip-compile-multi has been run after changing (multiple) requirements.in files

[pip-compile-multi](https://github.com/peterdemin/pip-compile-multi) allows to benefit from [pip-tools'](https://github.com/jazzband/pip-tools) dependency locking for requirements files combined with the possibility to break them down (eg. `base`, `dev`, `prod`, `test`)


asottile:
Thanks! :tada:

PR #212, opened 2019-02-05
Add Reference to `noeval`

brianjbuck:
Adds a hook to disallow `eval()` in code: 
-  https://github.com/brianjbuck/noeval

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -79,3 +79,4 @@
 - https://github.com/milin/giticket
 - https://github.com/sqlalchemyorg/zimports
 - https://github.com/peterdemin/pip-compile-multi
+- https://github.com/brianjbuck/noeval


asottile:
You might be interested in the [`pygrep`](https://pre-commit.com/#pygrep) hook type

For instance I think your hook python can be written instead as just metadata:

```yaml
-   id: noeval
    name: noeval
    entry: '\beval\('
    language: pygrep
    types: [python]
```

Might be a candidate for a PR to https://github.com/pre-commit/pygrep-hooks

brianjbuck:
Interesting. I didn't see that when I was looking...Let me investigate that route.

brianjbuck:
Closed in favor of using https://github.com/pre-commit/pygrep-hooks

PR #218, opened 2019-03-18
Adding in prometheus validation repo

fortman:
Adding in https://github.com/fortman/pre-commit-prometheus

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -82,3 +82,4 @@
 - https://github.com/sqlalchemyorg/zimports
 - https://github.com/peterdemin/pip-compile-multi
 - https://github.com/homebysix/pre-commit-macadmin
+- https://github.com/fortman/pre-commit-prometheus


asottile:
nice!  Does the builtin `docker_image` / `docker` types help out here at all? https://pre-commit.com/#docker_image

fortman:
@asottile There are two (maybe 3) options for docker image I did not see.  The first would be mounting a volume (--volume), and the second would be specifying a working directory inside the image (--workdir).

Also I wasn't 100% sure on how to override the docker CMD.  It needed to be updated to "check rules ${FILES}",  where files is the list of changed files to process.  I think that makes 3 things that the docker_image pre-commit didn't support.

asottile:
pre-commit will always add a volume that is the code directory being worked on and will set a workdir at the root of that (so it transparently looks like the executable is running from that location): https://github.com/pre-commit/pre-commit/blob/71c238d4ec61c0568949087f969bf9488663d578/pre_commit/languages/docker.py#L84-L85

You can add arguments to the command that runs by using `entry:`, they will be shlexed and put on the end

the `docker_image` docs go into a little more detail here: https://pre-commit.com/#docker_image

fortman:
That might indeed help me eliminate the shell scripts.  I'll take a look.  Thanks!

fortman:
@asottile I updated my repo to use docker_image.  Seems to be working :)

asottile:
sweet!

PR #226, opened 2019-04-21
Document prepare-commit-msg hooks

marcjay:
Update documentation to cover additional stage: prepare-commit-msg

(_new in X_ to be replaced with version number)

Accompanies https://github.com/pre-commit/pre-commit/pull/1004


asottile:
oops! that should be \*1\*.16.0

asottile:
oh you got it. 😆

PR #229, opened 2019-04-29
Update badges

asottile:


asottile:
lol, it turns out two thirds of these were referring to static images before this patch 😆

PR #232, opened 2019-05-02
ambv/black => python/black

asottile:
cc @ambv


ambv:
I'm sorry for adding to your workload!

asottile:
@ambv no problem :D -- happy to set up the new url

PR #234, opened 2019-05-25
Add the official hook from the golangci-lint project

syntaqx:
Both myself and @dnephin had implementations of golangci-lint - However, they've had an official one for [some time now](https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml) - I've sunset mine in my most recent version, and added this so that we can prevent duplication and work and try to get people to use & support the one from the project.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -39,6 +39,7 @@
 - https://github.com/Lucas-C/pre-commit-hooks-nodejs
 - https://github.com/Lucas-C/pre-commit-hooks-safety
 - https://github.com/chriskuehl/puppet-pre-commit-hooks
+- https://github.com/golangci/golangci-lint
 - https://github.com/dnephin/pre-commit-golang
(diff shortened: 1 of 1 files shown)


syntaqx:
@asottile If it was my own package I 100% would've put it at the bottom, but I figured this package deserved to be higher up since they're implementing the official version for something. Idk, but thanks for merging anyways, hah.

For the whole.. making it interacting/useful or whatever. What are you thinking? I might have some time and could play around with it.

asottile:
I haven't started working on it yet but basically what I want out of it is something like:

- grouping by language / target language
- ability to copy and paste configurations directly from the website
- star counts (?)
- some "common" configurations (maybe this'll  merge into `pre-commit sample-config` at some point, think like `pre-commit sample-config --black` or `pre-commit sample-config --asottile` etc. etc.

PR #235, opened 2019-06-05
Add cmake-format to the list

kveretennicov:


asottile:
looks like there's an open PR upstream as well -- maybe we can poke and get that merged instead :+1:

kveretennicov:
@asottile, I don't think they're very keen on accepting that PR.

This model of maintaining pre-commit hooks upstream is better suited for mature projects, with good compatibility story (or ones where all development stopped). At the moment, cmake-format is pre-1.0 and freely breaks formatting between releases. This is why my repo is pinned to 0.4.5, by the way.

asottile:
I don't think that's necessarily true -- pre-commit's design is for repeatability so you don't _need_ to pin, the consumers can choose the version in their `.pre-commit-config.yaml` -- for instance https://github.com/asottile/pyupgrade has changed formatting in almost every release, yet is widely used and works really well with `pre-commit`

asottile:
via #260 for now :( -- hopefully upstream will change their mind eventually

asottile:
thanks again for the patch!

PR #247, opened 2019-07-17
add R hooks

lorenzwalthert:
Adds R hooks from https://github.com/lorenzwalthert/pre-commit-hooks. Anything else I have to do?


asottile:
:bowing_man: thanks again for putting these together!

PR #248, opened 2019-07-25
add FelixSeptem hook

FelixSeptem:
add a set of golang's hook

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -96,3 +96,4 @@
 - https://github.com/marco-c/taskcluster_yml_validator
 - https://github.com/myint/docformatter
 - https://github.com/lorenzwalthert/pre-commit-hooks
+- https://github.com/FelixSeptem/pre-commit-golang


all-repos.yaml
```diff
@@ -96,3 +96,4 @@
 - https://github.com/marco-c/taskcluster_yml_validator
 - https://github.com/myint/docformatter
 - https://github.com/lorenzwalthert/pre-commit-hooks
+- https://github.com/FelixSeptem/pre-commit-golang
```
asottile: I notice you've included a few that are duplicated upstream -- for instance there's several other providers of `go-imports` already and `golangci-lint` has support for pre-commit in the official repo

FelixSeptem: the official `golangci-lint` will pending a lot of time in my macbook. I think it will install the whole golangci-lint (https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml), the already exist `goimports` one will overwrite local go file with param `goimports -w`,my `goimports` one will print the difference instead with `goimports -d` and filter `vendor` directory.

asottile: you should be able to configure the `args` to whatever you want

FelixSeptem: I chose the mild way to set the args, and the hook already work in some of my private repo.It looks like anything works fine till now, no pending,no chaos.I have already try some of the list one,not very suitable for my demand(some reason I said above).So I wanna to share my hooks with other who need this.If you think is duplicated.Close it if you like.

asottile: I'm just really not sure what this brings to the table that the others don't -- or why you wouldn't contribute to the others?

FelixSeptem: the first question: at least I provide a bazel buildifier(https://github.com/bazelbuild/buildtools/blob/master/buildifier/README.md) which is total new for the list(besides some adjust of args of command);the second question, I don't think a pull request flow will be faster than I write it in my own repo(and honestly I don't get the one which is convenient to get all hooks in one place for me to contribute), moreover I notice there already exist several `pre-commit-golang`, why not ask this when they are be listed?

asottile: I have no problem with this PR -- was just making sure you knew about the other options already in place

FelixSeptem: I make promise to keep update with this repo as possible.I use it in some of my repo,so I will make patch and process any valuable pull request as soon as possible

PR #250, opened 2019-07-30
Added clang-format to pre-commit-cpp

daverona:
Add the following hook to pre-commit-cpp:

* clang-format (default: Google C++ style)


asottile:
oh hmm, I thought I had put this on a cron loop so it picks up new hooks periodically -- let me just set that up (given this patch is empty!)

daverona:
Oh, thanks. (I wondered how to make newly added hooks show up in the supported hooks list after the repository registration has been done.)

asottile:
yep yep, it'll now get built once per day

PR #260, opened 2019-09-27
Add cmake-format to list of all hooks

iconmaster5326:
This MR adds [cmake-format-pre-commit-hook](https://github.com/iconmaster5326/cmake-format-pre-commit-hook) to the list of all hooks.

It is a hook to run [cmake-format](https://pypi.org/project/cmake-format/) on CMake files.

EDIT: I understand that there is a MR in the works in cmake-format itself to add this support. I heavily doubt that this MR will ever be merged, however, so I recommend accepting this MR over #235. It seems the owner of cmake-format isn't too keen on adding this hook, which is a shame.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -101,3 +101,5 @@
 - https://github.com/codingjoe/relint
 - https://github.com/nix-community/nixpkgs-fmt
 - https://github.com/d6e/beancount-check
+- https://github.com/iconmaster5326/cmake-format-pre-commit-hook
+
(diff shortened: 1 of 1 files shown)


asottile:
hmmm I would prefer we wait until the resolution of the other PR but I guess this is fine for now

I'd suggest regenerating your repository using [pre-commit-mirror-maker](https://github.com/pre-commit/pre-commit-mirror-maker) -- this'll have better tagging / versioning semantics -- plus it won't [unnecessarily depend on pyyaml](https://github.com/iconmaster5326/cmake-format-pre-commit-hook/blob/a53cfe56313f77c0b65470fb5d5c70af317d626e/setup.py#L12)

iconmaster5326:
Thanks for telling me about pre-commit-mirror-maker! I've gone and used that instead of my custom solution.

asottile:
ok, this seems fine enough to merge -- I'll keep an eye on the cmake-format PR (that would be _more_ ideal)

PR #264, opened 2019-11-01
Adding nbstripout to available hooks

ericmjl:
As per title.

Please let me know if I need to update anything else.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -105,3 +105,4 @@
 - https://github.com/thg-consulting/inspectortiger
 - https://gitlab.com/iamlikeme/nbhooks
 - https://github.com/Kuniwak/vint
+- https://github.com/kynan/nbstripout


asottile:
hooks should not invoke `git add`: https://github.com/kynan/nbstripout/blob/22f71272abcc719d84018e6abbd0a5f19eae6656/.pre-commit-hooks.yaml#L4

ericmjl:
Thanks for the note, @asottile.

Pinging @kynan on this thread for context. I enjoy using nbstripout, and I'd also like to see nbstripout on the official list of supported pre-commit hooks. Is the `git add` step necessary? I think conforming to the standard "don't git add" might be preferable here though.

kynan:
Yep, totally agree @ericmjl. We have a pending PR kynan/nbstripout#106 that will address this.

kynan:
FWIW, this PR needs rebasing :)

ericmjl:
Thanks @kynan! Updated! :smile:

kynan:
FYI, kynan/nbstripout#106 is now merged.

asottile:
sweet

PR #266, opened 2019-11-04
Add pre-commit hook for lisp-format

iconmaster5326:
lisp-format is a tool for autoformatting of Common Lisp code. I helped make the hook for that repo, so I thought I'd update the website here.


iconmaster5326:
Okay, I rebased. Hope that helps!

PR #270, opened 2019-11-17
Add gitlab-ci-linter

allburov:
We also have `check-gitlab-ci` in https://gitlab.com/smop/pre-commit-hooks, it's have external dependencies - `jq, curl`. 

New version `gitlab-ci-linter` is a pure-python script without external dependencies

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -109,3 +109,4 @@
 - https://github.com/ryanrhee/shellcheck-py
 - https://github.com/APIDevTools/swagger-cli
 - https://github.com/kynan/nbstripout
+- https://gitlab.com/devopshq/gitlab-ci-linter


asottile:
looks like this repository was deleted

allburov:
@asottile it's weird, I'll look at this, TY!

allburov:
I've recovered the repo, have no idea why it was deleted, we'll investigate that

asottile:
> I've recovered the repo, have no idea why it was deleted, we'll investigate that

sounds good!  you'll probably want to re-submit it to the `all-repos.yaml` list once the dust settles since I had to remove it to make the repo build (feel free to insert it into the same spot I deleted it from to retain the original ordering)

PR #272, opened 2019-12-04
Update H1 tag formatting to match other headings

dkolepp:
The `<h1>` tags are not generated with Markdown, while all other sub-headings are generated using Markdown.  Sub-headings render with a nice link, that make it easy for others to reference these sections (highlighted by red box in the example that follows):

## Before
---
image
---

## After
This PR causes `<h1>` headings to render the same as other sub-headings, creating links for subsections.  The following screenshot shows the same section rendered with the changes (green box):

---
image
---

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -18,7 +18,7 @@ from template_lib import md
     </div>
     <div class="col-sm-9">
         <div id="intro">
-            <div class="page-header"><h1>Introduction</h1></div>
+            <div class="page-header">${md('# Introduction')}</div>
 ${md('''
 Git hook scripts are useful for identifying simple issues before submission to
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -18,7 +18,7 @@ from template_lib import md
     </div>
     <div class="col-sm-9">
         <div id="intro">
-            <div class="page-header"><h1>Introduction</h1></div>
+            <div class="page-header">${md('''# Introduction''')}</div>
```
asottile: these don't need to be triple quoted strings since it's just one line

asottile:
idk why CI isn't working, shrugs shipit

dkolepp:
Thanks for the quick review and edit!

PR #277, opened 2019-12-14
Add blacklist hook

ahonnecke:
I'm not sure which branch I should be PRing against, but there seemed to be a merge conflict with `real_master`

This is the first pre-commit hook that I've written so apologies if I've done something silly.

This hook simple searches for the string "NOCOMMIT" and disallows that file (or hunk) to be committed.  I use this when I make a change that I know is temporary and should only exist locally to keep myself from committing that code.


asottile:
looks like you made your branch against `master` -- unfortunately due to the way that github pages works, that's the branch that gets published to pre-commit.com --  you'll want to rebase your branch on top of `real_master` -- that's where the source code is.  that branch gets built into the output which renders on the website

by the way, did you know that there's [`language: pygrep`](https://pre-commit.com#pygrep) which makes these types of "forbid X" hooks really simple!

ahonnecke:
I had not seen that, thank you!

PR #278, opened 2019-12-16
Add a very simple "Forbid string" hook

ahonnecke:


asottile:
ah ok, now that I see the code I don't think this is necessary (?) -- `pygrep` kinda does everything this hook can but better (more flexible, regex support, non-UTF8 files, etc.

also in the future please append to the bottom of the list :)

thanks for the PR nonetheless!

PR #297, opened 2020-02-04
Updated link from daverona-env/pre-commit-cpp to toscana/pre-commit-cpp

daverona:
My hook was registered already and I just changed the source repo of the hook.
Please merge.
Thanks a lot.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -98,7 +98,7 @@
 - https://github.com/myint/docformatter
 - https://github.com/lorenzwalthert/pre-commit-hooks
 - https://github.com/FelixSeptem/pre-commit-golang
-- https://gitlab.com/daverona-env/pre-commit-cpp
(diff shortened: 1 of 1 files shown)


arontier-egkim:
Hi, asottile 

From the last conversation with you (some three months ago?),
I remember you mentioning that a scheduled task wasn't running and that's why the hook page was not updated.  

Now I notice that the hook page is not updated this time again.
May be the same problem?  I just wanted to let you know about it.

Thanks.

asottile:
that's really weird! the page got updated in source but isn't rendered: https://github.com/pre-commit/pre-commit.github.io/commit/a8cbc8c90865ae49627ebec46a755eb4ddf6bbb3

asottile:
appears to have updated now... weird (I contacted support so maybe they kicked it)

PR #307, opened 2020-02-20
Add docs for post-checkout

andrewhare:
This adds docs for the new `post-checkout` hook.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -312,7 +312,7 @@ repository's configuration.
 =r=
     =c= `stages`
     =c= (optional) confines the hook to the `commit`, `merge-commit`, `push`,
-        `prepare-commit-msg`, `commit-msg`, or `manual` stage.  See
+        `prepare-commit-msg`, `commit-msg`, 'post-checkout', or `manual` stage.  See
         [Confining hooks to run at certain stages](#confining-hooks-to-run-at-certain-stages).
 =r=
     =c= `additional_dependencies`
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -1124,6 +1124,40 @@ the commit will be aborted.
 [commit-msg]: https://git-scm.com/docs/githooks#_commit_msg
 [prepare-commit-msg]: https://git-scm.com/docs/githooks#_prepare_commit_msg
 
+
+_new in __VERSION__: pre-commit can be used to manage [post-checkout] hooks.
```
asottile: this needs a header, probably something like `## pre-commit for switching branches`

the version will be 2.2.0 so you can amend that in right now

index.mako
```diff
@@ -1124,6 +1124,40 @@ the commit will be aborted.
...
+
+To use `post-checkout` hooks with pre-commit run:
+
+```console
+$ pre-commit install --hook-type post-checkout`
```
asottile: spurious backtick

index.mako
```diff
@@ -1124,6 +1124,40 @@ the commit will be aborted.
...
+or set working dir metadata properties. Since `post-checkout` doesn't operate
+on files, any hooks must set `always_run`:
+
+```yaml
+-   repo: local
+    hooks:
+    -   id: post-checkout-local
+        name: Post checkout
+        entry: bash -c ""
+        language: system
```
asottile: I would put `# ...` in place of the `language / entry` here since you're really only trying to highlight the `always_run` and `stages` options

index.mako
```diff
@@ -1125,12 +1125,13 @@ the commit will be aborted.
 [prepare-commit-msg]: https://git-scm.com/docs/githooks#_prepare_commit_msg
 
 
-_new in __VERSION__: pre-commit can be used to manage [post-checkout] hooks.
+## pre-commit for switching branches
+_new in 2.2.0: pre-commit can be used to manage [post-checkout] hooks.
```
asottile:
```suggestion
_new in 2.2.0_: pre-commit can be used to manage [post-checkout] hooks.
```

PR #308, opened 2020-02-22
add note about required arg for {prepare-}commit-msg

particledecay:
Since `--commit-msg-filename` is required when using prepare-commit-msg and commit-msg for things like `try-repo`, just making a note of that in the docs.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -488,6 +488,9 @@ directory while developing hooks.
 enabling a quick way to try out a repository.  Here's how one might work
 interactively:
 
+_note_: you may need to provide `--commit-msg-filename` when using this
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -1103,6 +1103,8 @@ pre-commit installed at .git/hooks/commit-msg
 current contents of the commit message which can be validated.  If a hook
 exits nonzero, the commit will be aborted.
 
+_note_: you may need to provide `--commit-msg-filename` when testing this hook type.
```
asottile: I think this probably makes more sense in the `pre-commit try-repo` / `pre-commit run` sections above

particledecay: Good point. I made this change.

PR #312, opened 2020-03-12
Add Pablo Iranzo Gómez repo for yaspeller

iranzo:
Add my repository for yaspeller hook

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -122,4 +122,4 @@
 - https://github.com/talos-systems/conform
 - https://github.com/twu/skjold
 - https://github.com/woile/commitizen
-- https://github.com/iranzo/precommit-hooks
+- https://github.com/hcodes/yaspeller.git
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -122,4 +122,4 @@
 - https://github.com/talos-systems/conform
 - https://github.com/twu/skjold
 - https://github.com/woile/commitizen
-- https://github.com/hcodes/yaspeller.git
+- https://github.com/hcodes/yaspeller
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -122,3 +122,4 @@
 - https://github.com/talos-systems/conform
 - https://github.com/twu/skjold
 - https://github.com/woile/commitizen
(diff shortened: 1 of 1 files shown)


asottile:
- it would be better to add the configuration to `yaspeller` directly
- worst case, it would be better to have a [mirror repository](https://github.com/pre-commit/pre-commit-mirror-maker) instead of a floating version

iranzo:
> * it would be better to add the configuration to `yaspeller` directly

The configuration can be done via .yaspeller.json directly, in fact the repository itself uses it with pre-commit hook

> * worst case, it would be better to have a [mirror repository](https://github.com/pre-commit/pre-commit-mirror-maker) instead of a floating version

I've been reading that repository but from the README it's not clear to me what is for.

It states `Scripts for creating mirror repositories that do not have .pre-commit-hooks.yaml` but then in the folder listing it shows the `.pre-commit-hooks.yaml`

We can work on that via an issue on the repo for the yaspeller hook, is that ok?

asottile:
no no, I mean the *pre-commit* configuration -- instead of your repo it would be best to contribute a `.pre-commit-hooks.yaml` to yaspeller's repository directly

iranzo:
Ah, got it, let me PR them

iranzo:
@asottile something like this? https://github.com/hcodes/yaspeller/pull/136

iranzo:
@asottile PR has been updated to point to yaspeller repository

all-repos.yaml
```diff
@@ -122,3 +122,4 @@
 - https://github.com/talos-systems/conform
 - https://github.com/twu/skjold
 - https://github.com/woile/commitizen
+- https://github.com/hcodes/yaspeller.git
```
asottile: none of the others end with .git -- why did you do it for this one?

iranzo: It was giving me a failure when I tested it , I can remove it and leave like the others

asottile: looks good but have to resolve the conflicts

iranzo: done!

mxr:
This was added in #344

PR #319, opened 2020-03-20
Give an example of how to force flake8 to use a specific python version.

Natim:
image

As flake8 documentation states, it might be interesting to tell people to set the version they want to target.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -1202,6 +1202,7 @@ as follows:
     hooks:
     -   id: flake8
         args: [--max-line-length=131]
+        language_version: python3.8
 ```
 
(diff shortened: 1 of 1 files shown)


asottile:
this is [already covered here](https://pre-commit.com/#overriding-language-version)  -- the section you've edited is completely unrelated

Natim:
> the section you've edited is completely unrelated

I agree with you it is unrelated to `language_version` however, it is related to flake8. I don't think it diminiss the example to add it there.

There are already unrelated lines that are related to flake8 only in this example.

asottile:
mmm no, the example is a minimal example to show `args` and only `args`

PR #320, opened 2020-03-21
Note pygrep UTF-8 limitation

scop:
Per discussion in https://github.com/pre-commit/pre-commit/pull/1322, I think the UTF-8 limitation deserves to be mentioned.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -749,6 +749,10 @@ way to write a simple hook which prevents commits by file matching.  Specify
 the regex as the `entry`.  The `entry` may be any python
 [regular expression](#regular-expressions).  For case insensitive regexes you
 can apply the `(?i)` flag as the start of your entry, or use `args: [-i]`.
+The regex operates in bytes mode and is internally encoded using UTF-8 for
+that, so the hook will produce correct results only for files in which the
+literal characters present in the regex are encoded the same as they are in
+UTF-8.
 
 _new in 1.8.0_: For multiline matches, use `args: [--multiline]`.
 


index.mako
```diff
@@ -749,6 +749,9 @@ way to write a simple hook which prevents commits by file matching.  Specify
 the regex as the `entry`.  The `entry` may be any python
 [regular expression](#regular-expressions).  For case insensitive regexes you
 can apply the `(?i)` flag as the start of your entry, or use `args: [-i]`.
+The regex operates in bytes mode and is internally encoded using UTF-8 for
+that, so the hook is guaranteed to produce correct results only against files
+encoded in UTF-8 and compatible encodings.
```
asottile: this is subtly incorrect.  for example an ascii regex will work with latin1 which is not UTF-8 compatible

scop: Oh, right, rephrased. It's not as concise as I'd like, but can't think of any better now.

asottile:
I'm still not quite sure how documenting this helps anyone or matters -- the pre-commit configuration *text files* must be UTF-8 also which would prevent anyone from even approaching the encoding of a pygrep hook's entry 🤔

scop:
The config file encoding and how the regex is encoded to a bytes one to are two separate things. The latter is done here: https://github.com/pre-commit/pre-commit/blob/fa8d02281373ec18c8463515c997291b6814e406/pre_commit/languages/pygrep.py#L75

If instead of `args.pattern.encode()` (which is equivalent to `...encode("UTF-8")` you'd have `args.pattern.encode("UTF-16BE")` there, the resulting regex would be applicable to UTF-16BE files, no matter what the config file charset is.

BTW I suppose we could actually add an arg specifying what charset to use to encode the regex to a bytes one, defaulting to UTF-8.

asottile:
I'm pretty sure nobody would actually care about this implementation detail enough for it to be prominently featured in the documentation.  You're right that `pygrep` *could* `.encode('UTF16-LE')` but that would be extremely unexpected behaviour, whereas sticking with UTF-8 is the expected choice.

As for the encoding parameter, unless someone actually demonstrates a usecase I see no reason to add the complexity

scop:
UTF-16BE was just an example demonstrating a flaw in the pre-commit text config file charset argument and its relation to the regex encoding. Surely UTF-16BE would not make sense as the default, UTF-8 is all good there.

Usecase for the charset arg: use pygrep to search for my surname in files encoded in ISO-8859-1.

asottile:
are you actually checking in files in ISO-8859-1?

scop:
I have a bunch of such files, yes. But no, I'm not actively pygrepping them. Others might have files in say UTF-16BE (just to use the same example) or some other encoding completely distinct from UTF-8 (e.g. some eastern Asian ones), and the pygrep docs as they stand give no hint that it will not be able to do the right thing with them.

asottile:
then I'm sorry but you don't have an actual use case

scop:
Sigh. And if I had said yes, I am actively pygrepping them, would you then have been willing to apply the patch in some form? I actually hope it would not have made a difference, because this is not at all about making pygrep work for some more or less imaginary case _for me_. To me this is about objective, technical correctness. It's trivial to see how this _will_ fail in valid, non-imaginary scenarios, it's easy enough to document the limitation, and it'd even be easy enough to also fix by providing a command line flag. But I guess we're done here.

asottile:
I understand your strive for correctness, but as far as I'm concerned if nobody is affected by it there's no reason to invest effort in building and maintaining a feature or to invest effort make the docs more verbose for something nobody is going to read or care about.  ~essentially KISS (though I don't really like the last S in the acronym)

PR #322, opened 2020-03-24
Add restructured text check

lorman:


asottile:
I would prefer you send a patch to the original repository

this mirror repository as written is also broken:
- it claims to be a mirror but doesn't utilize [pre-commit-mirror-maker](https://github.com/pre-commit/pre-commit-mirror-maker)
- it requires the user to install the tool globally
- it depends on  shell when it could be done directly in the configuration

please try upstream first, thanks!

lorman:
@asottile thanks! was wondering if you have some ideas around properly selecting the files to run the check on. Ran the mirror-maker with `--types rst` and got `'types': [rst]` in the .pre-commit-hooks.yaml, but when testing it doesn't seem to find any files to run against. Thanks!

asottile:
that should work? what repo did you try it on?

also I'd suggest trying to add the pre-commit configuration to https://github.com/myint/rstcheck directly first, that way we don't need to bounce through an intermediate repository and can send people directly to the tool!

PR #323, opened 2020-03-25
Add restructured text check

lorman:
Set up using pre-commit-mirror-maker.

Also opened a PR to add this config directly to rstcheck here: https://github.com/myint/rstcheck/pull/63


asottile:
via #342

PR #324, opened 2020-03-25
Yamlfmt hook

jumanjiman:


all-repos.yaml
```diff
@@ -30,6 +30,7 @@
 - https://github.com/elidupuis/mirrors-jscs
 - https://github.com/elidupuis/mirrors-sass-lint
 - https://github.com/jumanjihouse/pre-commit-hooks
+- https://github.com/jumanjihouse/yamlfmt
```
asottile: huh this is confusing, there's another tool called `yamlfmt` in python (and it's different from this one?): https://github.com/mmlb/yamlfmt

all-repos.yaml
```diff
@@ -64,6 +65,7 @@
 - https://github.com/psf/black
 - https://github.com/IamTheFij/ansible-pre-commit
 - https://github.com/IamTheFij/docker-pre-commit
+- https://github.com/macisamuele/language-formatters-pre-commit-hooks
```
asottile: these intentionally aren't alphabetized -- they're ordered by time-of-addition and then grouped by organization (so this should be added at the end)

all-repos.yaml
asottile: additionally, the package being added here clobbers the `pre_commit_hooks` python namespace (and should pick a different name/layout)

jumanjiman: should i rename my repo as `pre-commit-hook-yamlfmt` and update this?

do you mean that https://github.com/jumanjihouse/yamlfmt/blob/master/setup.py#L12 clobbers the namespace?
I can fix that as well!

all-repos.yaml
jumanjiman: GTK - i'll update

jumanjiman: pushed this fix to put at the end

all-repos.yaml
asottile: yep, both of those sound good

jumanjiman: updated

jumanjiman: from https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/blob/master/setup.py#L13

```
    # globally-unique name
    packages=[
        'jumanjihouse_precommit_yamlfmt',
    ],
```

jumanjiman:
tyvm!

PR #326, opened 2020-03-31
Fix incorrect sample code

revolter:


asottile:
this is correct as written, from [`language: fail`](https://pre-commit.com/#fail)

> The `entry` will be printed when the hook fails. It is suggested to provide a brief description for `name` and more verbose fix instructions in `entry`.

revolter:
Oh, wow, didn't know about `fail`, and I didn't pay attention.

PR #328, opened 2020-04-05
Include repo for Behave pre-commit hook :cucumber:

deepyaman:


asottile:
I don't think it makes sense to have `behave` as a pre-commit hook:
- it doesn't have access to your dependencies
- this is a fork of `behave` and not the actual behave

see also: the same rationale for "why not a pytest hook": https://github.com/pre-commit/pre-commit-hooks/issues/291#issuecomment-394167917

deepyaman:
> I don't think it makes sense to have `behave` as a pre-commit hook:
> 
> * it doesn't have access to your dependencies
> * this is a fork of `behave` and not the actual behave
> 
> see also: the same rationale for "why not a pytest hook": [pre-commit/pre-commit-hooks#291 (comment)](https://github.com/pre-commit/pre-commit-hooks/issues/291#issuecomment-394167917)

Thanks for your input; makes sense to me!

To the second point, I saw your similar point https://github.com/pre-commit/pre-commit.github.io/pull/312#issuecomment-598246726 after making my PR, and opened https://github.com/behave/behave/pull/824. I'll see if it get's accepted.

Issue #330, opened 2020-04-11
copyable code blocks for badges


asottile:
it would be nice to be able to copy and paste the code with a button or auto-select the code when clicking

ammarnajjar:
May I give this a shot? :)

asottile:
sure!

ammarnajjar:
@asottile , I thought that it might be a good idea to be able to copy any code block on the page, not just the badging section, so please give your take on the open PR #331. 
To make review easier, I made it of two commits: before and after javascript minification.

PR #331, opened 2020-04-13
add copy-icon top-left badges code blocks

ammarnajjar:
Resolves: #330


base.mako
```diff
@@ -59,6 +59,7 @@
         <footer class="pc-footer"><div class="container"></div></footer>
         <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
         <script src="node_modules/bootstrap-sass/assets/javascripts/bootstrap.min.js"></script>
+        <script>!function(){var e=document.getElementsByClassName("highlight");function t(e){var t=document.createElement("textarea");t.value=e,t.style.top="0",t.style.left="0",t.style.position="fixed",document.body.appendChild(t),t.focus(),t.select(),document.execCommand("copy"),document.body.removeChild(t)}for(let o of e){var n=document.createElement("img");n.setAttribute("src","./copy-icon.svg"),n.setAttribute("alt","copy"),n.setAttribute("title","Copy code"),n.setAttribute("width","16"),n.setAttribute("height","16"),o.insertBefore(n,o.children[0]),n.addEventListener("click",function(){t(o.getElementsByTagName("pre")[0].innerText)})}}();</script>
```
asottile, line 62: non-essential scripts should go at the end of body, also I'd rather not minify this in the source (it makes it much harder to modify)

copy-icon.svg
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
asottile, line 1: can you run this through `svgo`

ammarnajjar: done => 4e59e8a

base.mako
ammarnajjar: done => 02ae85e

copy-icon.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 247.744 247.744"><g fill="#000002"><path d="M206.89 0H74.824c-9.649 0-17.5 7.851-17.5 17.5v21.005h-16.47c-9.649 0-17.5 7.85-17.5 17.5v174.239c0 9.649 7.851 17.5 17.5 17.5h132.065c9.649 0 17.5-7.851 17.5-17.5V209.24h16.471c9.649 0 17.5-7.851 17.5-17.5V17.5c0-9.649-7.85-17.5-17.5-17.5zm-31.471 230.244c0 1.355-1.145 2.5-2.5 2.5H40.854c-1.355 0-2.5-1.145-2.5-2.5V56.005c0-1.356 1.145-2.5 2.5-2.5h132.065c1.355 0 2.5 1.144 2.5 2.5v174.239zm33.971-38.504c0 1.355-1.145 2.5-2.5 2.5h-16.471V56.005c0-9.65-7.851-17.5-17.5-17.5H72.324V17.5c0-1.355 1.144-2.5 2.5-2.5H206.89c1.355 0 2.5 1.145 2.5 2.5v174.24z"/><path d="M149.553 77.121H64.22a7.499 7.499 0 00-7.5 7.5c0 4.143 3.357 7.5 7.5 7.5h85.333c4.143 0 7.5-3.357 7.5-7.5a7.5 7.5 0 00-7.5-7.5zM149.553 110.898H64.22c-4.143 0-7.5 3.357-7.5 7.5s3.357 7.5 7.5 7.5h85.333c4.143 0 7.5-3.357 7.5-7.5s-3.357-7.5-7.5-7.5zM149.553 144.676H64.22a7.499 7.499 0 00-7.5 7.5c0 4.143 3.357 7.5 7.5 7.5h85.333c4.143 0 7.5-3.357 7.5-7.5s-3.357-7.5-7.5-7.5zM149.553 178.453H64.22a7.499 7.499 0 00-7.5 7.5c0 4.143 3.357 7.5 7.5 7.5h85.333c4.143 0 7.5-3.357 7.5-7.5a7.5 7.5 0 00-7.5-7.5z"/></g></svg>
```
asottile, line 1: this should live in the assets folder

base.mako
```diff
@@ -68,5 +68,35 @@
...
+        btn.setAttribute("title", "Copy code");
+        btn.setAttribute("width", "16");
+        btn.setAttribute("height", "16");
+        block.insertBefore(btn, block.children[0]);
+        btn.addEventListener("click", function() {
+          var text = block.getElementsByTagName('pre')[0].innerText;
+          copyTextToClipboard(text);
+        })
+  }
+})()
```
asottile, line 99: this script should live in the assets folder, the indentation here is also inconsistent

base.mako
```diff
@@ -68,5 +68,35 @@
...
+    textArea.style.left = "0";
+    textArea.style.position = "fixed";
+    document.body.appendChild(textArea);
+    textArea.focus();
+    textArea.select();
+    document.execCommand('copy');
+    document.body.removeChild(textArea);
+  }
+  var codeBlockElements = document.querySelectorAll('div#advanced ul')[8].getElementsByClassName("highlight");
+  for (let block of codeBlockElements) {
```
asottile, line 86: I believe `for (let ... of ...) {` is less portable -- caniuse says only 93%: https://caniuse.com/#feat=mdn-javascript_statements_for_of

base.mako
```diff
@@ -68,5 +68,35 @@
...
             ga('send', 'pageview');
         </script>
+        <script>
+(function () {
+  function copyTextToClipboard(text) {
+    var textArea = document.createElement("textarea");
+    textArea.value = text;
+    textArea.style.top = "0";
+    textArea.style.left = "0";
+    textArea.style.position = "fixed";
```
asottile, line 78: is it necessary to style this? could we use css here -- could this be placed offscreen (I believe this blinks in and out right now if I'm reading this correctly)

base.mako
```diff
@@ -68,5 +68,35 @@
...
+    document.body.appendChild(textArea);
+    textArea.focus();
+    textArea.select();
+    document.execCommand('copy');
+    document.body.removeChild(textArea);
+  }
+  var codeBlockElements = document.querySelectorAll('div#advanced ul')[8].getElementsByClassName("highlight");
```
asottile, line 85: this seems very fragile -- perhaps we can add a specific class to the section containing this?

base.mako
```diff
@@ -68,5 +68,35 @@
...
+  for (let block of codeBlockElements) {
+        var btn = document.createElement("img");
+        btn.setAttribute("src", "./copy-icon.svg");
+        btn.setAttribute("alt", "copy");
+        btn.setAttribute("title", "Copy code");
```
asottile, line 90: maybe "copy to clipoard"?

base.mako
ammarnajjar: - style is used to prevent scrolling when the element is added to the dom, but as I am testing it now, only `position` seem to be needed
- it does not work with css being defined somewhere else. My understanding for the reasoning is that this piece of script does not know about styles defined outside, and if it is desired to be used, one would have to import it somehow.
- right, it blinks for a tiny moment, but after using `position` only (as I tried it) it won't.

asottile: `position: absolute; left: -9999px; top: -9999px;` might also be something to consider?

base.mako
ammarnajjar: - agree
- first thought is to add some `html` to the template and add whatever needed (id or class), do you think there is a better way doing it?

asottile: we already do some custom stuff to render some of the code blocks -- could probably just adjust that [here](https://github.com/pre-commit/pre-commit.github.io/blob/bff157f8a1f8651e1b9089280c8eef36648ffaca/template_lib.py#L102-L108)

maybe something like

```markdown

    ```markdown  # copyable
    ...
    ```
```

base.mako
ammarnajjar: - unfortunately, any position other than `fixed` will affect the scrolling, so I think that there is no run away from using it
- I used a negative number for `top` and `left` though, so that the textarea is created off-screen.

base.mako
ammarnajjar: WIP

base.mako
ammarnajjar: done.

base.mako
ammarnajjar: I updated the script to use the normal `for(var i=0; ....; i++)`

base.mako
ammarnajjar: a new folder with name "assets" is created and the script is moved there

copy-icon.svg
ammarnajjar: moved to assets folder

template_lib.py
```diff
@@ -3,10 +3,12 @@
 import shlex
 import subprocess
 import sys
+from typing import Any
+from typing import Type
 
 import markdown_code_blocks
 import markupsafe
-
+import mistune
```
asottile, line 11: this is not what I had imagined -- it should be something much simpler like this:

```python
copyable_s = '  # copyable'
copyable = lang.endswith(copyable_s)
lang, _, _ = lang.rpartition(copyable_s)
...

return f'<div class="copyable">{ret}</div>'
```

ammarnajjar: - The regex for parsing `lang` lives [here](https://github.com/lepture/mistune/blob/2837628008dbe1d8f01e76845fcb00c157e28667/mistune.py#L112), and in order to be able to catch ` # copyable`, I thought that this regex ought to be overridden, that led making custom `Lexer` and `Grammar` classes.
- I might be missing the spot where your example here works, can you please direct me there?

asottile: `def block_code` -- you already edited it :)

ammarnajjar: - also, I thought about adding additional `div` around `ret`, but then I switched to using the original `div` to keep the structure consistent 😄

ammarnajjar: - sorry if I argue, but I test with this minimal sample: https://pastebin.com/10UaNLeY , and `lang` is never parsed if it does not match the regex defined [here](https://github.com/lepture/mistune/blob/2837628008dbe1d8f01e76845fcb00c157e28667/mistune.py#L112), and as a result the `block_code` function is not called 🤔

asottile: then how about `rst#copyable` ?

ammarnajjar: - `rst#copyable` sure works, I tool your [example](https://github.com/pre-commit/pre-commit.github.io/pull/331#discussion_r407777535) so literally 😄 
- I will re-do it to accept such `rst#copyable` lang

ammarnajjar: - btw, it might be worth mentioning that the current implementation supports not only `copyable` class but also any other class, so if you needed to add a print button to specific code blocks, it will be as easy as adding `printable` class (for example), and react to it in a javscript code.
- and to make the story even funnier, I wanted to go further and support multiple classes i.e: ` # copyable printable .. etc` on the same line, but I thought I should get your feedback first, and glad I did 😆

asottile: yagni though :)

ammarnajjar: done!

index.mako
```diff
@@ -1501,27 +1501,27 @@ you use pre-commit!
...
+  ```rst#copyable
   .. image:: https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit&logoColor=white
      :target: https://github.com/pre-commit/pre-commit
      :alt: pre-commit
   ```
 
 - AsciiDoc:
 
-  ```
+  ``` #copyable
```
asottile, line 1524: I think this one needs the space removed iirc?

scss/main.scss
```diff
@@ -202,6 +202,10 @@ h1,
     .s2 { color: #8ae234; }
 }
 
+.copyable {
+    img { cursor: pointer; }
```
asottile, line 206: let's put this on its own line like the others -- scss-lint has an exclusion of this, but it's only intended for the pygments styles which were copy pasted

asottile: this can also be written without nesting as just `.copyable img {`

template_lib.py
```diff
@@ -100,12 +100,22 @@ def header(self, text: str, level: int, raw: str) -> str:
         )
 
     def block_code(self, code: str, lang: str) -> str:
+        copyable = False
+        if lang:
+            copyable_s = '#copyable'
+            copyable = lang.endswith(copyable_s)
+            if copyable:
```
asottile, line 107: oh, you shouldn't need this if statment -- can switch to `partition` and it'll do what you want

base.mako
```diff
@@ -68,5 +68,6 @@
             ga('create', 'UA-104682927-1', 'auto');
             ga('send', 'pageview');
         </script>
+        <script src="assets/addCopyToBadging.js" type="text/javascript"></script>
```
asottile, line 71: let's just call this `copyable.js` -- also I think in html5 the `type="text/javascript"` is assumed and can be removed

index.mako
ammarnajjar: it's optional, but I removed it

base.mako
ammarnajjar: done

template_lib.py
ammarnajjar: 🤦  done

scss/main.scss
ammarnajjar: done

ammarnajjar:
- Thanks for bearing up with me @asottile 
- I noticed that the icon does not show on https://pre-commit.com/, the class `copyable` is there though 🤔

asottile:
ah, assets need to be whitelisted in deploy, lemme fix that: d0abd9741fb0a9a3c04a5306520846cd353f500d

PR #343, opened 2020-05-17
Add GitLab caching example

mattcanty:


index.mako
```diff
@@ -1579,6 +1579,17 @@ cache:
   - $HOME/.cache/pre-commit
 ```
 
+### gitlab example
+
+```yaml
+variables:
+  XDG_CACHE_HOME: "$CI_PROJECT_DIR/.cache/"
```
asottile, line 1586: I don't think this is correct, this puts the pre-commit cache inside the clone of the repository (which is likely to have negative side-effects)

mattcanty: OK. I'm actively investigating this right now.

mattcanty: Just following on from a conversation you've had earlier. Will see what effect this has on our repo.
https://github.com/pre-commit/pre-commit/issues/1207

PR #353, opened 2020-05-21
Add the Nitpick development tool

andreoliwa:
Add the pre-commit hook for [andreoliwa/nitpick: Enforce the same configuration (flake8, isort, mypy, pylint) across multiple Python projects](https://github.com/andreoliwa/nitpick).

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -134,3 +134,4 @@
 - https://github.com/myint/rstcheck
 - https://github.com/hcodes/yaspeller
 - https://github.com/maltzj/google-style-precommit-hook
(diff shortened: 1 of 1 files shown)


asottile:
usually the suggested execution for flake8 plugins is to use flake8 + `additional_dependencies` [for example](https://github.com/asottile/flake8-2020#as-a-pre-commit-hook)

andreoliwa:
Oh, I see...
Thanks, I will change that. 🙂 

So I don't even need `.pre-commit-hooks.yaml` on my project, right?

asottile:
indeed!

PR #356, opened 2020-06-07
Updates index.mako

sumit4613:
I guess the latest versions in the documentation shall be updated to their latest stable versions.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -113,13 +113,13 @@ pre-commit --version
 ```yaml
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v2.4.0
     hooks:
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -113,13 +113,13 @@ pre-commit --version
 ```yaml
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v2.4.0
```
asottile, line 116: this isn't the latest version

sumit4613:
@asottile i didn't knew that these versions were generated, thanks for letting me know. 🙂

asottile:
they're not yet -- but it would be better if they were instead of many manual PRs

sumit4613:
You're right, I'll try to implement this.

PR #361, opened 2020-06-19
Add support for named anchors for all configuration keys for direct linking

marcjay:
Add special handling for anchors that match the format: `[foo](_#bar)` that adds a named anchor before a self-link
Replace documentation references to configuration keys with links to key documentation where possible

This is the result of some yak-shaving when trying to write some documentation as per https://github.com/pre-commit/pre-commit-hooks/issues/500#issuecomment-646272468
I was hoping to be able to link directly to the relevant configuration key but couldn't do so. Appreciate this is non-standard markdown, so am very open to a better approach, but hopefully this would be a useful addition when pointing users at the docs

### DIFF
--- a/template_lib.py
+++ b/template_lib.py
@@ -16,6 +16,7 @@ SPECIAL_CHARS_RE = re.compile('[^a-z0-9 _-]')
 ROW = '=r='
 COL = '    =c= '
 INDENT = ' ' * 8
+SELF_LINK_PREFIX = '_#'
 
 
 def _render_table(code: str) -> str:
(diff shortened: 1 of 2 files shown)


template_lib.py
```diff
@@ -88,6 +89,19 @@ def _render_cmd(code: str) -> str:
 
 
 class Renderer(markdown_code_blocks.CodeRenderer):
+    def link(
+        self, link: str, text: Optional[str], title: Optional[str],
+    ) -> str:
+        match = NAMED_ANCHOR_RE.search(link)
+        if match:
```
asottile, line 96: can probably just use `.startswith('_#')`

template_lib.py
```diff
@@ -88,6 +89,19 @@ def _render_cmd(code: str) -> str:
...
+            a_id = match.group(1)
+            return (
+                f'<a name="{a_id}"></a>'
+                f'<a href="#{a_id}">{title}</a>'
```
asottile, line 100: does `<a name="{...} href="#{...}">...</a>` work?

template_lib.py
marcjay: Ah yes, at one point the regex was going to be more complicated but no longer, have replaced with that

template_lib.py
marcjay: Turns out it does! I was copying the approach from Terraform documentation ([example](https://www.terraform.io/docs/providers/aws/r/instance.html#cpu_core_count)) where they have them separated, but yup, it works combined - have changed it

PR #364, opened 2020-06-26
Adds a caching example for CircleCI

remeika:
Updates docs about managing CI caches to include general guidance on how to version the cache, and a specific example of how to do this in CircleCI.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -1646,6 +1646,33 @@ my_job:
       - ${PRE_COMMIT_HOME}
 ```
 
+### circleci example
+
+like [azure pipelines](#azure-pipelines-example), circleci also uses
+immutable caches:
+
+```yaml
+  steps:
+  - run:
+    command: |
+      cp .pre-commit-config.yaml pre-commit-cache-key.txt
+      python --version --version >> pre-commit-cache-key.txt
+  - restore_cache:
+    keys:
+    - v1-pc-cache-{{ checksum "pre-commit-cache-key.txt" }}
+
+  # ...
+
+  - save_cache:
+    key: v1-pc-cache-{{ checksum "pre-commit-cache-key.txt" }}
+    paths:
+      - ~/.cache/pre-commit
+```
+
+(source: [@chriselison])
+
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -1633,6 +1637,28 @@ immutable caches:
...
+### circleci example
+
+like [azure pipelines](#azure-pipelines-example), circleci also uses
+immutable caches:
+
+```yaml
+steps:
+    - run:
+        command: |
+            cat .pre-commit-config.yaml > pre-commit-deps.txt
```
asottile, line 1649: UUOC: `cp` does this better

but why make a separate file, can't you just use the original file?

also looks like `pre-commit-deps.txt` is unused

index.mako
```diff
@@ -1572,6 +1572,10 @@ which may be faster, use something like
 
 [XDG Base Directory Specification]: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
 
+If your CI environment uses immutable caches, the pre-commit repository should
+be versioned using both the contents of your configuration file (typically
+`.pre-commit-config.yaml`) and the output of the command `python -VV`.
```
asottile, line 1577: this isn't always completely true, but is a good estimate for the `healthy()` check

index.mako
```diff
@@ -1633,6 +1637,28 @@ immutable caches:
...
 
+### circleci example
+
+like [azure pipelines](#azure-pipelines-example), circleci also uses
+immutable caches:
+
+```yaml
+steps:
+    - run:
+        command: |
```
asottile, line 1648: this indentation doesn't line up properly (either use 2 or 4 space indents, but not mixed)

index.mako
remeika: Good info. Added a note below with that caveat.

index.mako
remeika: This was a copy-pasta error. Creating a new file, `pre-commit-deps.txt`, to concatenate `.pre-commit-config.yaml` & `python -VV`, since CircleCI can't use environment variables in cache key templates.

index.mako
remeika: Fixed.

PR #369, opened 2020-07-23
docs: add swiftformat hook

levibostian:
I created one pre-commit hook for [swiftformat](https://github.com/nicklockwood/SwiftFormat). This PR is to add the repo to the site for other swift devs to find.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -136,3 +136,4 @@
 - https://github.com/jvstein/pre-commit-dotnet-format
 # - https://github.com/hakancelik96/unimport
 - https://github.com/PeterMosmans/jenkinslint
+- https://github.com/nicklockwood/SwiftFormat


levibostian:
Not sure why the CI build failed? From the error logs, I can't see the problem.

asottile:
ah, it looks like you linked your own repo instead of the swiftformat one, perhaps that's why it's failing?  I'm not sure why there's no output of that though 🤔

asottile:
not sure why the error didn't show up, there was a hook which incorrectly set `verbose: true` (I've since fixed it)

levibostian:
🤦......

I can't believe that I am *just now* realizing that the [swiftformat](https://github.com/nicklockwood/SwiftFormat/blob/master/.pre-commit-hooks.yaml) project has it's own hooks file.......

Yesterday was my very first day using pre-commit so I guess beginner's mistake. I made my own repo for noting! But, hey. I learned a lot and had a pleasant experience working with pre-commit. Bonus: The official repo got added to the site. 

Thanks for catching that.

asottile:
no problem!  thanks for the PR 🎉

PR #370, opened 2020-07-26
Add graphene-django-precommit-hook

jackton1:


all-repos.yaml
```diff
@@ -137,3 +137,4 @@
 # - https://github.com/hakancelik96/unimport
 - https://github.com/PeterMosmans/jenkinslint
 - https://github.com/nicklockwood/SwiftFormat
+- https://github.com/tj-pre-commit/graphene-django-precommit-hook
```
asottile, line 140: fwiw this part is unnecessary: https://github.com/tj-pre-commit/graphene-django-precommit-hook/blob/219e5e363913b57ef7a0eb572f2a33928c09f3d0/entrypoint.sh#L3-L7

you also don't need a setup.py / setup.cfg since it's just a script, `language: script` is probably better

this probably also fits better as a [repository local hook](https://pre-commit.com/#repository-local-hooks) as I don't believe `manage.py graphql_schema` is a standard entrypoint for django (and therefore unlikely to be generically reusable)

thoughts?

jackton1: Yeah, I agree this would be useful as part of the graphene django repo IMO

Issue #375, opened 2020-08-10
what is the base directory for "log_file" output #1553


drjasonharrison:
Originally reported as https://github.com/pre-commit/pre-commit/issues/1553

I'm running pytest using the following definition:

```
-   repo: local
    hooks:
    -   id: simple-pytest
        description: run pytest
        name: simple-pytest
        entry: pytest
        language: system
        types: [python]
        pass_filenames: false
        stages: [push]
        log_file: pytest.log
```
and I see pre-commit list pytest during the git push after I make a change to a .py source file and git add and git commit. But I cannot find a pytest.log file anywhere.

When I run pytest manually:
```
pytest > manual-pytest.log
```
The manual-pytest.log file is created in the current working directory.

This is on Ubuntu 18.04.4, with python 3.6.9, pytest 5.4.3

Conclusion: pytest is not failing -- that is why log file is created.

asottile:
no need to make a separate issue, you can use `closes pre-commit/pre-commit#123` to close across projects

PR #376, opened 2020-08-10
#375: expand description for log_file

drjasonharrison:
Expanded the description of log_file to include the details that it is created only when the hook fails. Also added an example of redirecting the output of a hook for when you want the output of a hook copied to a log file regardless of the success/failure.

Also noticed that when italics of a version number ends the spacing after the version number is too narrow. So I added `&thinsp;&nbsp;` after the end of italics.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /build
 /node_modules
 /venv
+package-lock.json
--- a/index.mako
+++ b/index.mako
@@ -181,7 +181,7 @@ pre-commit config file describes what repositories and hooks are installed.
 
 ## .pre-commit-config.yaml - top level
 
-_new in 1.0.0_: The default configuration file top-level was changed from a
+_new in 1.0.0_ : The default configuration file top-level was changed from a
 list to a map.  If you're using an old version of pre-commit, the top-level
 list is the same as the value of [`repos`](#pre-commit-configyaml---repos).
 If you'd like to migrate to the new configuration format, run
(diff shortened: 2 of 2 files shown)


index.mako
```diff
@@ -181,7 +181,7 @@ pre-commit config file describes what repositories and hooks are installed.
 
 ## .pre-commit-config.yaml - top level
 
-_new in 1.0.0_: The default configuration file top-level was changed from a
+_new in 1.0.0_&thinsp;:&nbsp;: The default configuration file top-level was changed from a
```
asottile, line 184: please don't touch these, thanks

.gitignore
```diff
@@ -6,3 +6,4 @@
 /build
 /node_modules
 /venv
+package-lock.json
```
asottile, line 9: please leave gitignore alone, thanks

index.mako
```diff
@@ -330,8 +330,10 @@ repository's configuration.
         the hook passes.  _new in 1.6.0_.
 =r=
     =c= [`log_file`](_#config-log_file)
-    =c= (optional) if present, the hook output will additionally be written
-        to a file.
```
asottile, line 334: instead of a comma splice how about `if present, the hook output will additionally be written to a file when the hook fails or [verbose](...) is true`

index.mako
```diff
@@ -1282,14 +1284,15 @@ A `local` hook must define [`id`](#hooks-id), [`name`](#hooks-name), [`language`
...
+Here's an example configuration with a few `local` hooks. Note how the pylint
+hook redirects it's output to a log file for possible examination:
 
 ```yaml
 -   repo: local
     hooks:
     -   id: pylint
         name: pylint
-        entry: pylint
+        entry: pylint |& tee -a pylint.log
```
asottile, line 1295: this isn't correct: there's no shells involved in pre-commit's execution

PR #381, opened 2020-08-15
Add pre-commit hooks for Open Policy Agent (OPA)

anderseknert:
Not sure if anything else is needed. If so, let me know :)

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -139,3 +139,4 @@
 - https://github.com/nicklockwood/SwiftFormat
 - https://github.com/hukkinj1/mdformat
 - https://gitlab.com/daverona/pre-commit/php
+- https://github.com/anderseknert/pre-commit-opa


all-repos.yaml
```diff
@@ -139,3 +139,4 @@
 - https://github.com/nicklockwood/SwiftFormat
 - https://github.com/hukkinj1/mdformat
 - https://gitlab.com/daverona/pre-commit/php
+- https://github.com/anderseknert/pre-commit-opa
```
asottile, line 142: just a few tips:

I think you can replace all of the python code in the project with simple `system` configuration

for instance here's `opa fmt`:

```yaml
- id: opa-fmt
  name: OPA fmt
  description: Formats Rego policy using opa fmt
  entry: opa fmt
  args: [-w]
  language: system
  files: (\.rego)$
```

anderseknert: Huh, I didn’t realize that! That’s cleaner indeed. I’ve some plans for a couple of hooks that will be more advanced, so likely will require python there, but I guess they can be mixed freely? If so I’ll update my repo tomorrow.

asottile: yep then can be mixed freely, I'd also suggest not using a module named `pre_commit_hooks` since that'll class with https://pypi.org/p/pre-commit-hooks if they were ever installed together somewhow

anderseknert: Thanks @asottile 👍 will rename that with the other change then and ping you once I’ve done so.

anderseknert: Suggestions applied @asottile. Again, thanks for pointing them out to me.

PR #382, opened 2020-08-16
Add radix-ai/auto-smart-commit

lsorber:
This PR:
- ~~Sorts all hooks alphabetically to help improve browsability on https://pre-commit.com/hooks.html. The only exception to this rule is that all of the pre-commit organisation's hooks are listed first.~~
- Adds a new hook https://github.com/radix-ai/auto-smart-commit that automatically formats commit messages as [Jira smart commits](https://confluence.atlassian.com/fisheye/using-smart-commits-960155400.html).

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -140,3 +140,4 @@
 - https://github.com/hukkinj1/mdformat
 - https://gitlab.com/daverona/pre-commit/php
 - https://github.com/anderseknert/pre-commit-opa
+- https://github.com/radix-ai/auto-smart-commit


asottile:
they are intentionally not sorted, please revert that

lsorber:
@asottile Undid the sorting, and added the new hook on the last line. What is the convention if I may ask?

asottile:
they are ordered to give early adopters preferential position

all-repos.yaml
```diff
@@ -140,3 +140,4 @@
 - https://github.com/hukkinj1/mdformat
 - https://gitlab.com/daverona/pre-commit/php
 - https://github.com/anderseknert/pre-commit-opa
+- https://github.com/radix-ai/auto-smart-commit
```
asottile, line 143: I'd suggest removing the suggestion of `default_stages` from your example configuration, maybe suggest it but don't require it (it doesn't always do what someone would want)

lsorber: Removed that line from the README.

asottile:
> > they are ordered to give early adopters preferential position
> 
> As a pre-commit user, I don't think sorting by "oldest first" is a very useful way to sort and discover all the hooks.
> 
> In my opinion, sorting by "newest first", "author", and "alphabetically" seem the most useful, perhaps combined with a few "pinned" hooks or a "recommended" category that should receive preferential treatment.
> 
> Also, it's not clear to me which insert date to use: the date the hook was published on GitHub, or the date the hook was inserted in this list?

well sorting by github user isn't any more helpful

it's time inserted into the list of course, the other is untenable

PR #387, opened 2020-09-02
Improve doc on autoupdate CLI options

l3d00m:
- Make clearer what the default behaviour really is
- Remove a leftover closing bracket
- Make clear it uses latest commit because bleeding edge might be ambiguous. (Is that even correct that it uses latest commit? Didn't know what else bleeding edge could mean.)

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -798,9 +798,9 @@ Auto-update pre-commit config to the latest repos' versions.
 
 Options:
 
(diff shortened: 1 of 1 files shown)


asottile:
no improvement whatsoever

l3d00m:
Huh, sorry to hear that, I thought it would be better to understand like this. Nevertheless thanks for the great tool, I really appreciate it.

PR #391, opened 2020-09-22
Add boidolr/pre-commit-hooks to list of pre-commit hook repos

boidolr:


asottile:
is there a way you could name this repository to differentiate it? it might get confused with pre-commit/pre-commit-hooks

additionally some of these seem like they'd be better contributed to https://github.com/pre-commit/pygrep-hooks or are duplicates of others that already exist

boidolr:
As these are already used by this point I am not going to change the name anymore (also it's not the only repo with that name - I get that you'd not want to add more though).
I am open to adding the pygrep ones to https://github.com/pre-commit/pygrep-hooks though my impression was that the hooks there were geared towards python specifically.

asottile:
renaming a repo is easy and github will still allow you to clone at the old url -- for example ambv/black and python/black both go to psf/black

asottile:
pygrep-hooks is meant to be generic, note the "Naming Scheme" heading (it just happens that most of them have been geared towards python so far)

PR #393, opened 2020-09-23
Adds cmake-format-precommit to list of all_repos

Skylion007:
Adds the official cmake-format-precommit hook to the list of repos.

Should the other cmake-format hook be removed from the all_repos.yaml hooks? It's more or less the same but is not the official one. I elected to leave it for now.


asottile:
yeah if this one is official we should remove the other one

Skylion007:
@asottile Done

asottile (approved):

PR #396, opened 2020-09-25
Use the "literal" block style instead of the "folded" one

glasserc:
Verbose regular expressions allow comments, but if we use "folded"
multiline strings, then the comment will also comment out the
remainder of the regular expression.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -1477,7 +1477,7 @@ the `(?x)` regex flag.
 ```yaml
 # ...
     -   id: my-hook
-        exclude: >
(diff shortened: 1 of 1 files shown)


glasserc:
Still looking at this... I guess the existing example works, and it's because "more indented" lines have newlines preserved. I think this is an odd behavior and since we're feeding the whole thing to the regular expression engine, we still may as well leave in the newlines.

asottile (approved):
yaml sucks I'm sorry

PR #398, opened 2020-09-30
Add JS Beautify to list of repos

aufdenpunkt:


asottile:
have you proposed this upstream? it would be preferable there instead of a mirror repository.  I'm inclined to decline this and suggest an attempt be made there first

aufdenpunkt:
I'm not sure if I got you right. Is it about the pre-commit hook itself? Or because I have forked the pre-commit.com repo to add the link of the pre-commit hook repo? What do you mean with "proposed this upstream"?

asottile:
for example, adding the `.pre-commit-hooks.yaml` directly to the repository of the tool you're wrapping -- no need to have a mirror repository (that'll be out-of-date / not-locked / etc.).

aufdenpunkt:
Aaah, okay. Makes absolute sense to me. Thanks for clarifying 👍 => Closed

PR #399, opened 2020-10-03
Add pycln to the list of hook repos.

hadialqattan:
I've added a hook for finding and removing unused import statements from Python file(s).


asottile (approved):
just so you know there's also `unimport` and `autoflake` which have the same idea

hadialqattan:
> just so you know there's also `unimport` and `autoflake` which have the same idea
> 
> 

Yeah, but each tool is designed differently with different features.

PR #401, opened 2020-10-07
Explanation about local hooks

erfannariman:
background, see [SO](https://stackoverflow.com/questions/64036351/pre-commit-for-local-hook-gives-error-unrecognized-arguments-pre-commit-conf)

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -444,7 +444,7 @@ file that tells pre-commit:
         the hook passes.  _new in 1.6.0_.
 =r=
     =c= [`pass_filenames`](_#hooks-pass_filenames)
-    =c= (optional: default `true`) if `false` no arguments will be passed to
+    =c= (optional: default `true`) if `false` no filenames will be passed to
         the hook.
 =r=
     =c= [`require_serial`](_#hooks-require_serial)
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -445,7 +445,8 @@ file that tells pre-commit:
 =r=
     =c= [`pass_filenames`](_#hooks-pass_filenames)
     =c= (optional: default `true`) if `false` no arguments will be passed to
-        the hook.
+        the hook. Set this to `True` when creating a local hook and passing arguments through
+        [`entry`](#hooks-entry).
```
asottile, line 449: this isn't quite right, it has nothing to do with that -- it's only about passing filenames along

erfannariman: I see, maybe what I tried to achieve here is to be more explicit. Right now "no arguments will be passed" is quite confusing especially since there's an `args` argument.

What about "no files names will be passed as arguments to the hook"?

asottile: I would just replace the `arguments` -> `filenames` in the existing messaging

erfannariman: Sure, done.

asottile (approved):

PR #403, opened 2020-10-09
Make it clear where config file should go

lfdebrux:
It took me a while to work out where the file `.pre-commit-config.yaml` in step 2 should go. I think this small change will make it clearer for others.

### DIFF
--- a/index.mako
+++ b/index.mako
@@ -100,9 +100,9 @@ conda install -c conda-forge pre-commit
 pre-commit --version
 ```
 
-### 2. Add a pre-commit configuration
+### 2. Add a pre-commit configuration to your project
 
-- create a file named `.pre-commit-config.yaml`
+- create a file named `.pre-commit-config.yaml` in the root level of your project
 - you can generate a very basic configuration using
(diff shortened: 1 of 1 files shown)


index.mako
```diff
@@ -100,9 +100,9 @@ conda install -c conda-forge pre-commit
 pre-commit --version
 ```
 
-### 2. Add a pre-commit configuration
+### 2. Add a pre-commit configuration to your project
```
asottile, line 103: I don't think this adds anything

index.mako
```diff
@@ -100,9 +100,9 @@ conda install -c conda-forge pre-commit
...
 
-- create a file named `.pre-commit-config.yaml`
+- create a file named `.pre-commit-config.yaml` in the root level of your project
```
asottile, line 105: this is already mentioned [here](https://pre-commit.com/#adding-pre-commit-plugins-to-your-project)

asottile (changes requested):
the word "project" is ambiguous, I'd rather rely on the existing documentation

thanks for the patch but I'm going to decline this

PR #405, opened 2020-10-27
Added hooks for Fireblade

Firebladee:


asottile:
https://github.com/shellcheck-py/shellcheck-py is a better implementation of this that's already in the list

Firebladee:
Unfortunately that still needs shellcheck installed on your machine.  One of the hooks I have is to have shellcheck done through docker.
I can remove the normal shellcheck hook if you want.  I only add the normal shellcheck hook as completeness.

asottile:
it does not, try it

Firebladee:
Before I wrote my hook I looked at that and dismissed it as I thought I needed one running through docker.
Oh well.  Thanks for creating pre-commit catches so much stuff.

asottile:
yep yep -- that's why @ryanrhee and I made it -- because shellcheck is so useful and also tricky to install :)

PR #410, opened 2020-11-02
Add add-msg-issue-preffix-hook to supported (prepare-commit-msg)

avilaton:
The hook searches the current branch for something looking like a jira issue and prepends the commit message.
It is a prepare-commit-msg hook.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -147,3 +147,4 @@
 - https://github.com/hadialqattan/pycln
 - https://github.com/nbQA-dev/nbQA
 - https://github.com/Scony/godot-gdscript-toolkit
+- https://github.com/avilaton/add-msg-issue-prefix-hook


asottile:
neat, note that `subprocess.getoutput` is usually considered problematic (as it invokes a shell) -- I'd suggest using `check_output` instead

also note that this:

```python
   except Exception as e:
        print(e)
        pass
```

the `pass` line can be removed as it does nothing (pass is mostly meant to be a placeholder where you need a block to exist, but the bloc already exists due to the `print(e)`)

here's a video I did on `pass`: https://youtu.be/ANjDsGZZXy8?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY

asottile:
also `always_run` is not needed -- prepare-commit-msg hooks will always be run anyway since there's no file filtering

avilaton:
Thanks @asottile ! sorry you saw those print(e) and the rest, we had this as a raw manual hook and migrated to a pre-commit hook in a rush, we will clean it up and add some customization options. Pre-commit rocks, thanks for making and maintaining it.

PR #411, opened 2020-11-02
add new repo for JVM hooks

dustinsand:
Currently contains one hook to run the Detekt static analyzer on Kotlin source files.


asottile:
not sure if you saw but the latest version of pre-commit adds support for the `coursier` package manager which can install jvm tools -- might be something to try as well!

PR #416, opened 2020-11-09
Add gitown

milin:
https://github.com/milin/gitown


asottile (approved):
a few fixes to your README:
- don't suggest installing such an old version of pre-commit
- [don't suggest `rev: master`](https://pre-commit.com/#using-the-latest-version-for-a-repository) -- suggest a tag instead
- the README mentions commit-msg, though the hook isn't a commit-msg hook afaict

PR #417, opened 2020-11-12
mention license on main page

taleinat:
As things are, the only way to see the license for pre-commit is on the project's GitHub page or in the source.


asottile:
I don't think this improves anything

PR #421, opened 2020-11-19
Add pipx as installation option

reegnz:


asottile:
this list isn't intended to be exhaustive and I don't think `pipx` provides anything over `pip` here

reegnz:
I can only quote the pipx docs:

> pip is a general-purpose package installer for both libraries and apps with no environment isolation. pipx is made specifically for application installation, as it adds isolation yet still makes the apps available in your shell: pipx creates an isolated environment for each application and its associated packages.

So the key here is isolation.

I'm sorry that you don't see value in that.

asottile:
> 
> 
> I can only quote the pipx docs:
> 
> > pip is a general-purpose package installer for both libraries and apps with no environment isolation. pipx is made specifically for application installation, as it adds isolation yet still makes the apps available in your shell: pipx creates an isolated environment for each application and its associated packages.
> 
> So the key here is isolation.
> 
> I'm sorry that you don't see value in that.

it's not that I don't see value, it's that anything that works for pip also works for pipx so there's no point in putting that in the docs here

cs01:
> it's not that I don't see value, it's that anything that works for pip also works for pipx so there's no point in putting that in the docs here

I don't really follow this. There *are* things that work in pipx that don't work in pip. pip installs all packages together and can silently introduce conflicting packages, whereas pipx isolates each package and its dependencies.

asottile:
> > it's not that I don't see value, it's that anything that works for pip also works for pipx so there's no point in putting that in the docs here
> 
> I don't really follow this. There _are_ things that work in pipx that don't work in pip. pip installs all packages together and can silently introduce conflicting packages, whereas pipx isolates each package and its dependencies. I understand not wanting to make the list exhaustive, just wanted to mention that.

you're agreeing with me :)

gaborbernat:
I'd just like to point out that users more often than not will use what you listed, and by only listing pip you're essentially encouraging users to install pre-commit at the system level, which can cause dependency conflict issues... At the very least ideally should guide users to hey install it in a virtual environment, please... 🤷

asottile:
modern `pip` defaults to `--user` installs when not run as root which is no different to what [`pipx`'s installation instructions are](https://pipxproject.github.io/pipx/installation/#install-pipx)

admittedly if I had my way, I wouldn't even suggest that -- I'd suggest `curl https://bootstrap.pypa.io/virtualenv.pyz -o virtualenv.pyz && python3 virtualenv.pyz venv && rm virtualenv.pyz && venv/bin/pip install pre-commit` but let's be realistic

either way, I think we all can agree that it's a waste of everyone's time to add unnecessary documentation for pipx, poetry, pipenv, faster-than-pip, pip-tools, etc. etc. etc. ways to install pre-commit which all essentially boil down to "it's a python package"

gaborbernat:
> "it's a python package"

Perhaps python.org needs to have a how to install a python package we can all link to.

hexagonrecursion:
> I don't think `pipx` provides anything over `pip`

`pipx` does add something: it is a _significantly_ more convenient method of installing and managing CLI *applications* than `pip`.

<table>
<tr><td>pipx</td><td>pip</td></tr>
<tr><td>

```sh
pipx run pre-commit
```

</td>
<td>

```sh
# Oops: forgot to install
$ pre-commit
bash: pre-commit: command not found
# Install
$ pip install pre-commit
# Run
$ pre-commit
```

</td></tr>
<tr><td> Installs every tool into a separate virtual environment and manages the environments for me </td><td> I have to *learn* how to manually create and manage virtual environments. This is especially painful for users who do not actually use Python themselves. </td></tr>
<tr><td>keeps the tool up to date</td><td>updates have to be installed manually</td></tr>
</table>

asottile:
please don't gravedig two year old resolved issues -- thanks

PR #423, opened 2020-11-20
Add some Dart/Flutter related hooks to `all-repos.yaml`

lig:
FYI, @Cretezy, @dluksza, @Matt-Gleich.


asottile:
@Cretezy: btw this can be `language: system` and then `entry: flutter format` instead of needing a separate script
@dluksza: same advice as well!

it would probably be cool to add native dark support to pre-commit at some point, so that pre-commit could manage these installs instead of needing the system to install them, but that can be done later

asottile (approved):

PR #432, opened 2020-12-01
Fix GitHub actions example

atugushev:
Use `set-output` instead of `set-env`, since it is [removed](https://github.blog/changelog/2020-11-09-github-actions-removing-set-env-and-add-path-commands-on-november-16/) from GHA.

Addresses https://github.com/jazzband/pip-tools/pull/1012#discussion_r419011371


index.mako
```diff
@@ -1728,12 +1728,13 @@ like [azure pipelines](#azure-pipelines-example), github actions also uses
 immutable caches:
 
 ```yaml
-    - name: set PY
-      run: echo "::set-env name=PY::$(python -VV | sha256sum | cut -d' ' -f1)"
```
asottile, line 1732: a less intrusive change is `>> $GITHUB_ENV`

asottile (approved):

index.mako
atugushev: Yeah, updated in a25c534

PR #435, opened 2020-12-19
added id and type filter for hooks

### LINKED ISSUE #341: Group hooks by type
It would be really handy to have the hooks grouped by what it specified in the `types` field in the JSON data. That way if I'm looking to see what hooks apply to Python I can look in that section. I would assume this would be its own `<h2>` section after the current by-repository list.

paulhfischer:
resolves #341 

- options for select are automatically generated and only include types for hooks that are in `all-hooks.json`
- also added a search input for the hook id
- repos with non-matching hooks are completely hidden

### DIFF
--- a/assets/filter_repos.js
+++ b/assets/filter_repos.js
@@ -1,0 +1,41 @@
+'use strict';
+
+(() => {
+    const searchInput = document.getElementById('search-hook-id');
+    const selectInput = document.getElementById('select-hook-type');
+
+    const hooks = document.getElementById('hooks');
+    const repos = hooks.getElementsByTagName('ul');
+
+    const filterHooks = () => {
+        const id = searchInput.value.toLowerCase();
+        const type = selectInput.value;
+
+        for (let i = 0; i < repos.length; i += 1) {
+            const repo = repos[i];
+            let hasVisibleHooks = false;
+            const repoHooks = repo.getElementsByTagName('li');
+
+            if (repoHooks) {
+                for (let j = 0; j < repoHooks.length; j += 1) {
+                    const repoHook = repoHooks[j];
+                    const hookId = repoHook.dataset.id.toLowerCase();
+                    const hookTypes = repoHook.dataset.types;
+
+                    if (hookId.includes(id) && hookTypes.includes(type)) {
+                        repoHook.hidden = false;
+                        hasVisibleHooks = true;
+                    } else {
+                        repoHook.hidden = true;
+                    }
+                }
+            }
+
+            repo.hidden = !hasVisibleHooks;
+            hooks.querySelector(`h3[data-repo="${repo.dataset.repo}"]`).hidden = !hasVisibleHooks;
+        }
+    };
+
+    searchInput.addEventListener('input', filterHooks);
+    selectInput.addEventListener('change', filterHooks);
+})();
--- a/hooks.mako
+++ b/hooks.mako
@@ -3,6 +3,22 @@
 
 <div class="page-header"><h1>Supported hooks</h1></div>
 
+<div class="row">
+    <div class="col-xs-6 form-group">
+        <label for="search-hook-id">Filter hook id:</label>
+        <input id="search-hook-id" type="text" placeholder="filter hook id" class="form-control"></input>
+    </div>
+    <div class="col-xs-6 form-group">
+        <label for="search-hook-id">Filter hook type:</label>
+        <select id="select-hook-type" class="form-control">
+            <option value="">--</option>
+            % for hook_type in sorted(all_types, key=str.lower):
+                <option value="${hook_type}">${hook_type}</option>
+            % endfor
+        </select>
+    </div>
+</div>
+
 <p>
 To add to this list, fork <a href="https://github.com/pre-commit/pre-commit.com">this repository</a>.
 </p>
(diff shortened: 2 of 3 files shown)


hooks.mako
```diff
@@ -1,7 +1,18 @@
...
-<div class="page-header"><h1>Supported hooks</h1></div>
+<div class="page-header">
+    <h1>Supported hooks</h1>
+    <div class="filter">
+        <input id="search-hook-id" type="text" placeholder="filter hook id"></input>
+        <select id="select-hook-type">
+            <option value="">filter hook type</option>
+            % for hook_type in all_types:
+                <option value="${hook_type}">${hook_type}</option>
+            % endfor
```
asottile, line 12: these should be sorted

hooks.mako
```diff
@@ -1,7 +1,18 @@
 ## -*- coding: utf-8 -*-
 <%inherit file="base.mako" />
 
-<div class="page-header"><h1>Supported hooks</h1></div>
+<div class="page-header">
+    <h1>Supported hooks</h1>
+    <div class="filter">
+        <input id="search-hook-id" type="text" placeholder="filter hook id"></input>
```
asottile, line 7: for the smallest responsive size, this needs more margin

I wonder if there's a bootstrap forms thing that does all this for you

assets/filter_repos.js
```diff
@@ -0,0 +1,37 @@
...
+            }
+        }
+
+        repo.hidden = !hasVisibleHooks;
+        hooks.querySelector(`h3[data-repo="${repo.dataset.repo}"]`).hidden = !hasVisibleHooks;
+    }
+};
+
+searchInput.addEventListener('input', filterHooks);
+selectInput.addEventListener('change', filterHooks);
```
asottile, line 37: will need to look at caniuse for:
- `let`
- fat arrow
- format strings
- `const`
- `.includes(...)`
- `.dataset`

I suspect at least one of those things will need to be made more-compatible

assets/filter_repos.js
```diff
@@ -0,0 +1,37 @@
+const searchInput = document.getElementById('search-hook-id');
+const selectInput = document.getElementById('select-hook-type');
+
+const hooks = document.getElementById('hooks');
+const repos = hooks.getElementsByTagName('ul');
```
asottile, line 5: this sets all of these as globals -- this should be wrapped in an anonymous function (+ `'use strict';`)

hooks.mako
```diff
@@ -1,7 +1,18 @@
...
+        <select id="select-hook-type">
```
asottile, line 8: for some reason this select looks really broken on windows

would it be better to have labels for both of these inputs and then have the placeholder value here be `--` or something?  might make this a little nicer

make_templates.py
```diff
@@ -25,7 +25,12 @@ def get_env() -> Dict[str, Any]:
         open('all-hooks.json').read(),
         object_pairs_hook=collections.OrderedDict,
     )
-    return {'all_hooks': all_hooks}
+    all_types = {
+        hook_type
+        for properties in all_hooks.values()
+        for hook_type in properties[0].get("types", [])
```
asottile, line 31: there's also `types_or` now, not sure how this will filter 🤔

scss/main.scss
```diff
@@ -225,3 +225,31 @@ h3 {
...
+    select {
+        padding: 6px 12px;
+        box-sizing: border-box;
+        height: 38px;
+    }
+
+    input {
+        width: 250px;
+    }
+}
```
asottile, line 255: will need to caniuse for:

- flexbox
- justify-content: space-between
- :not(...)
- :last-child (though iirc this has wide support)

scss/main.scss
```diff
@@ -225,3 +225,31 @@ h3 {
         display: inline;
     }
 }
+
+.page-header {
```
asottile, line 229: I think most of this css can be replaced with the bootstrap classes probably

assets/filter_repos.js
paulhfischer: the whole code fits the ES6, detailed caniuse:

* `let`: [96.75%](https://caniuse.com/let)
* fat arrow: [95.47%](https://caniuse.com/arrow-functions)
* format strings: [95.99%](https://caniuse.com/template-literals)
* `const`: [99.16%](https://caniuse.com/const)
* `.includes(...)`: [95.98%](https://caniuse.com/es6-string-includes)
* `.dataset`: [99.54%](https://caniuse.com/dataset)

to meet `>0.2%, not dead`, `const` and `let` have to be changed to var and the template literal with e.g. `.concat`.

hooks.mako
paulhfischer: i restructured the header with bootstrap `col` and `row`.

hooks.mako
paulhfischer: i think this should be fixed as i now use the bootstrap classes for styling the inputs.

added labels and moved the inputs out of the header, as the header would look really cluttered imo (and too high compared to the other sites due to the labels).

make_templates.py
paulhfischer: i joined `types` and `types_or` now.

scss/main.scss
paulhfischer: i was able to get rid of all those items as i moved the inputs out of the header.

scss/main.scss
paulhfischer: so the problem is that the site currently depends on bootstrap 3, which doesn't support the `d-flex` etc. classes. i can update it to bootstrap 4, this however causes an error when being compiled with `pysassc` (thus, we might have to switch to another compiler).

let me know if i should do this update, for now i renamed the classes to match the naming of the needed bootstrap 4 classes.

for now i renamed the flexbox classes to match the naming of the bootstrap 4 ones.

asottile: it isn't a problem with the compiler, there's just a bunch of breaking changes between 3 and 4 (I tried migrating at some point).  bs4 works fine with `pysassc` (it's the same as node-sass) -- I maintain it btw

paulhfischer:
(i will squash the commits later, but it might be easier to see what i changed with a separate commit)

scss/main.scss
paulhfischer: ah, okay, didn't look at the error message in detail tbh (as it's also possible with bs3). the error was reported in line 2 which is the bootstrap import, so i guessed the error was caused by bootstrap itself (but it's not).

asottile (approved):

PR #440, opened 2020-12-25
updated to bootstrap 4

paulhfischer:
as bootstrap 4 is EOL since [07/19](https://github.com/twbs/release#release-schedule), i thought it might make sense to update the bootstrap version (including some breaking changes). short summery of what i changed:

- updated jquery and bootstrap to current version
- font (`Source Sans Pro`) is added via `@import` in the `scss`-file instead of a link in the `head` of `index.html`
- navigation now uses `svg` instead of `png`-version of the logo
- updated github-button according to [buttons.github.io](https://buttons.github.io/)
- some css classes were replaced by bootstrap 4 classes.
- current page is highlighted in navigation


scss/main.scss
```diff
@@ -1,150 +1,92 @@
...
 }
 
-pre {
-    margin-bottom: 20px;
+footer {
+    // scss-lint:disable ImportantRule
+    position: absolute  !important;
+    right: 0;
+    bottom: 0;
+    left: 0;
```
asottile, line 20: I think the fix for this is to remove padding-bottom on `body`?

asottile (commented):
- Responsive tables are broken
- The h1s are missing their underline (this was done with custom css before)

scss/main.scss
paulhfischer: with the absolute positioning (in combination with `bottom: 0`) the footer is sticky (noticable if you search for hooks and the content height becomes lower than your browser height)

asottile: Ah I see, that is better -- I didn't consider the hooks page

paulhfischer:
> * Responsive tables are broken
> * The h1s are missing their underline (this was done with custom css before)

thanks, `page-header` was a class in v3 that has been removed in v4, but i added it manually to the css :)

scss/main.scss
```diff
@@ -1,150 +1,120 @@
...
+}
 
-    h1 {
-        font-weight: bold;
-        margin: 40px 0 20px;
 
-        a {
-            color: $pc-yellow;
+// underline headline
+.page-header {
```
asottile, line 45: It looks like the suggested way to do this is with the helper classes?

https://stackoverflow.com/a/49708022/812183

paulhfischer: mhm yes i can do it that way, but imo the custom class might be cleaner (and more understandable) in the html (and as we are using scss, the border color etc. is inherited from the theme)

asottile: Ah ok cool, sounds good

asottile (approved):

paulhfischer:
mhm are you using a cache for the css files? seems like it isn't refreshing on pre-commit.com …

asottile:
Yeah there's two levels of caching that are outside my control -- One is the github pages cache and the other is cloudflare

I'm pushing a hacky fix to master which should "solve" that issue

PR #441, opened 2020-12-31
Add check-untracked-migrations hook

ecugol:
Hello @asottile !

This PR adds [check-untracked-migrations](https://github.com/ecugol/pre-commit-hooks-django) hook.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -158,3 +158,4 @@
 - https://gitlab.com/pablodiehl/pre-commit-lua-formatter
 - https://github.com/frnmst/md-toc
 - https://github.com/mgedmin/check-manifest
+- https://github.com/ecugol/pre-commit-hooks-django


asottile:
a few things to improve:

- I'd recommend using a different namespace than `pre-commit-hooks` + `pre_commit_hooks` as those are already in use by the official project pre-commit/pre-commit-hooks
- your hook currently doesn't work quite right.  I believe you want to change the following in your `.pre-commit-hooks.yaml`:

(note: `type:` isn't a thing, and you want your hook to always run independent of the files changed because it detects non-staged things)

```diff
 -   id: check-untracked-migrations
     name: Untracked Django migrations checker
     description: "Forbid untracked Django migrations"
     entry: check-untracked-migrations
     language: python
-     type: [python]
+    always_run: true
+    pass_filenames: false
```

- this code does nothing, you can remove it: https://github.com/ecugol/pre-commit-hooks/blob/0990453d6cab4c9b1fa1f6eb82aacddae74ea721/pre_commit_hooks/check_untracked_migrations.py#L17-L18

ecugol:
Hello @asottile, thank you very much for the tips. I've modified the namespace as well and added args handling to specify on which branches hook will work. 

I hope it is ok now.

asottile:
I would change your namespace to `pre_commit_hooks_django` to match your package name personally

this should probably return 0 if I understand your `--branches` argument correctly ?

https://github.com/ecugol/pre-commit-hooks-django/blob/29fac94aa9fb4a2045e8defe59009bd730364748/hooks/check_untracked_migrations.py#L17-L19

asottile (approved):

PR #443, opened 2021-01-07
Add blocklint to list of hook repositories

troycomi:
Blocklint is a python-based linter for checking any text or source code file for the usage of blocklisted words such as master/slave.  The list of words and sensitivity to case and word boundaries can be configured but the defaults are meant to promote inclusive language in source code.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -159,3 +159,4 @@
 - https://github.com/frnmst/md-toc
 - https://github.com/mgedmin/check-manifest
 - https://github.com/ecugol/pre-commit-hooks-django
+- https://github.com/PrincetonUniversity/blocklint


asottile:
you probably want `types: [text]` otherwise your hook will be applied to [binary files](https://github.com/PrincetonUniversity/blocklint/blob/42e97b15550e5178d6b2d40444b058f2502b04f9/.pre-commit-hooks.yaml#L6)

this tool might be replaceable with just [pygrep](https://pre-commit.com/#pygrep), then you wouldn't need to maintain any code beyond a list of patterns

troycomi:
Good catch on the types!  

Pygrep is a valid alternative.  Blocklint only has base python requirements and outputs issues similar to a compiler for use in linters.  Words can be configured to ignore special characters, which I think makes it easier to search for invalid words in source code.  For instance, setting blocklist to 'cat' gives the regex
```
c[^a-zA-Z0-9]?a[^a-zA-Z0-9]?t  # with re.IGNORECASE
```
This allows targets like `blocklist` to match `myBlockList`, `block_list` or `block-LIST`.

asottile:
interesting, it's been shown that such a set of patterns is ineffective, for example [mars explorer](http://www.mit.edu/activities/safe/labelling/0198f1.html)

troycomi:
Very true, and this certainly isn't supposed to be used for censoring.  The use case is developers that want to remove uses of master/slave, whitelist/blacklist and maybe a handful of non-inclusive words or maybe disallowed libraries.

If they want to get around the restrictions, they don't have to install it in the first place.  False positives can be handled by allowing a maximum number of issues through or marking the line with an ignore pragma.

Thanks again, this is a new project and I appreciate the feedback!

PR #447, opened 2021-01-15
Add sondrelg/pep585-upgrade to all hooks

sondrelg:
Hi! 

I've just finished writing my first real pre-commit hook, and I thought I would add it to the list 🎉

Any feedback on the implementation would also be welcome if you have any 👏

Here's the repo: https://github.com/sondrelg/pep585-upgrade

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -162,3 +162,4 @@
 - https://github.com/PrincetonUniversity/blocklint
 - https://github.com/sirosen/check-jsonschema
 - https://github.com/sirosen/fix-smartquotes
+- https://github.com/sondrelg/pep585-upgrade


asottile:
I would have preferred a contribution to https://github.com/asottile/pyupgrade/issues/343

regex replacing is going to be pretty error prone, you'd be better to use the tokenization as pyupgrade does

asottile:
as for generic feedback:
- you probably don't want to set `stages: [commit]` as that limits the usefulness of your tool
- the general convention for executable names is to not use underscores (`_`) but instead use dashes (`-`)

sondrelg:
> - you probably don't want to set stages: [commit] as that limits the usefulness of your tool
> - the general convention for executable names is to not use underscores (_) but instead use dashes (-)

Thanks! Removed and updated 👍

> regex replacing is going to be pretty error prone, you'd be better to use the tokenization as pyupgrade does

Yep, so I get all imports and annotations using `ast`, then only use regex to find the replacement candidate within a single line. Hadn't heard of `tokenization` until now, but I'll definitely read up on it.

-------------------

What do you think about the project structure? 

It took me a while to figure out a structure where I didn't need to put everything in one file, but I see now that most pre-commit hooks _do_ put all code into one file.

asottile:
> 
> 
> > * you probably don't want to set stages: [commit] as that limits the usefulness of your tool
> > * the general convention for executable names is to not use underscores (_) but instead use dashes (-)
> 
> Thanks! Removed and updated 👍
> 
> > regex replacing is going to be pretty error prone, you'd be better to use the tokenization as pyupgrade does
> 
> Yep, so I get all imports and annotations using `ast`, then only use regex to find the replacement candidate within a single line. Hadn't heard of `tokenization` until now, but I'll definitely read up on it.
> 
> What do you think about the project structure?
> 
> It took me a while to figure out a structure where I didn't need to put everything in one file, but I see now that most pre-commit hooks _do_ put all code into one file.

structure seems fine -- I usually find that a single file thing for very simple things works better for me (though, some have grown out of control like pyupgrade 😆 -- I need to split that at some point)

 A few small things on the structure though:
- probably your top level module should be the same as your package
- probably want a `__main__.py` so someone can `python -m pep585_upgrade` as an alternative to run your script

PR #448, opened 2021-01-16
Register hook for vulture

neutrinoceros:
Register hooks from https://github.com/jendrikseipp/vulture

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -163,3 +163,4 @@
 - https://github.com/sirosen/check-jsonschema
 - https://github.com/sirosen/fix-smartquotes
 - https://github.com/sondrelg/pep585-upgrade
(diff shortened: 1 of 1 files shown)


all-repos.yaml
```diff
@@ -163,3 +163,5 @@
 - https://github.com/sirosen/check-jsonschema
 - https://github.com/sirosen/fix-smartquotes
 - https://github.com/sondrelg/pep585-upgrade
+- https://github.com/ikamensh/flynt
```
asottile, line 166: sorry, the flynt project is not welcome here -- past conduct of the maintainer is not appropriate

neutrinoceros: Oh. I didn't know. I'll stick to vulture then.

asottile (approved):

PR #450, opened 2021-01-20
Add no-string-hints

MarcoGorelli:
Hi,

Adding a hook for [no-string-hints](https://github.com/MarcoGorelli/no-string-hints)


asottile (approved):
neat!

PR #451, opened 2021-01-21
Add jupytext to list of hooks

Skylion007:
* We recently added official pre-commit hook support to jupytext, a program which updates Python notebooks to/from text file representations.

Look forward to seeing this included in the official list of hooks @asottile 😁


asottile:
this is the default so it shouldn't be in the configuration: https://github.com/mwouts/jupytext/blob/41b7e9c93b50da1d6feb486aceb2c4d534374090/.pre-commit-hooks.yaml#L6

asottile (approved):

JohnPaton:
Ah good catch, I had been experimenting with it and obviously forgot to remove the line, nice!

PR #454, opened 2021-01-26
added tflint pre-commit from source.

jtcarnes:
Added tflint pre-commit from the creators of tflint. Uses golang to install, which is different than the ones currently in the list which expect the binary.


asottile (approved):
awesome!  I think I'll start using this myself now :)

PR #455, opened 2021-01-27
Add hook for protolint

wwuck:


all-repos.yaml
```diff
@@ -168,3 +168,4 @@
 - https://github.com/mwouts/jupytext
 - https://github.com/ejba/pre-commit-maven
 - https://github.com/terraform-linters/tflint
+- https://github.com/yoheimuta/protolint
```
asottile, line 171: it would probably be better if the source repository also allowed for installation via `go` (since pre-commit has native support for `go`)

I think that would look like this:

```yaml
# .pre-commit-hooks.yaml
---
- id: protolint
  name: Lint Protocol Buffer Files
  description: Runs protolint to lint Protocol Buffer files
  language: go
  types: ["proto"]
  entry: protolint lint
- id: protolint-docker
  name: Lint Protocol Buffer Files
  description: Runs protolint Docker image to lint Protocol Buffer files
  language: docker_image
  types: ["proto"]
  entry: yoheimuta/protolint lint
```

wwuck: Thanks, i'll submit a PR to protolint.

Would this mean the `.pre-commit-config.yaml` needs updating to specify the language, or does pre-commit choose automatically depending on which tool (go, docker) is installed?  What would happen if both go and docker were installed for this hook and `language` is not specified in `.pre-commit-config.yaml`?

asottile: nothing special is needed from the configuration side, both can be installed side-by-side even -- the config would look like:

```yaml
# in .pre-commit-config.yaml
-    repo: https://github.com/yoheimuta/protolint
    rev: ... # pick a sha / rev
    hooks:
    -   id: protolint
    # -   id: protolint-docker (use this one if you have docker available but not go)
```

in the future, pre-commit might even be able to bootstrap a go-based environment from nothing as it does for ruby / node, but that hasn't been implemented yet

wwuck: Ahh thanks, I missed the change in `id:`.

wwuck:
Checks should be fixed after https://github.com/yoheimuta/protolint/pull/150 is merged.

asottile:
> 
> 
> Checks should be fixed after [yoheimuta/protolint#150](https://github.com/yoheimuta/protolint/pull/150) is merged.

oof sorry I misled you there!

asottile (approved):

PR #457, opened 2021-01-28
Add tfsec from source

jtcarnes:
Added TFSec pre-commit from TFSec repo. Different than the current tfsec hooks as this builds the package directly from go instead of expecting it to be pre-built on the system.


asottile (approved):
awesome!

PR #458, opened 2021-02-01
Add official hadolint repository hook

wwuck:
There was already an existing hook listed for hadolint, but the official hadolint repository has now added their own pre-commit hooks.

https://github.com/hadolint/hadolint/pull/523
https://github.com/hadolint/hadolint/pull/527

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -25,6 +25,7 @@
 - https://github.com/asottile/dead
 - https://github.com/asottile/setup-cfg-fmt
 - https://github.com/asottile/cheetah_lint
+- https://github.com/hadolint/hadolint
 - https://github.com/digitalpulp/pre-commit-php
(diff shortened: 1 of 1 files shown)


all-repos.yaml
```diff
@@ -25,6 +25,7 @@
 - https://github.com/asottile/dead
 - https://github.com/asottile/setup-cfg-fmt
 - https://github.com/asottile/cheetah_lint
+- https://github.com/hadolint/hadolint
```
asottile, line 28: the list is in order by appearance so this should go to the end

wwuck: Ok.  I didn't want people to get confused and use the older third-party hook implementation when there is now one on the official repository.

I'll update the order to move to the end.

asottile (approved):

PR #463, opened 2021-02-08
fix paragraph spacing in tables

asottile:


paulhfischer:
ah good catch, seems like i only looked at the tables with one paragraph and missed the ones with multiple `p`-tags / `pre`-when i added this selector …

PR #466, opened 2021-02-13
abs-imports

MarcoGorelli:
A little tool to automatically convert relative imports to absolute ones

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -174,3 +174,4 @@
 - https://github.com/google/go-jsonnet
 - https://github.com/guid-empty/flutter-dependency-validation-pre-commit
 - https://github.com/cpplint/cpplint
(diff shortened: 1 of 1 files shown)


asottile:
just a tip from posix conventions -- executables are usually dashes instead of underscores: https://github.com/MarcoGorelli/abs-imports/blob/5554d5c65b65ecb1b62d697b06f5e53838ede3f7/setup.cfg#L28

asottile (approved):

MarcoGorelli:
> just a tip from posix conventions -- executables are usually dashes instead of underscores: https://github.com/MarcoGorelli/abs-imports/blob/5554d5c65b65ecb1b62d697b06f5e53838ede3f7/setup.cfg#L28

Thanks for the tip, really appreciate it!

PR #475, opened 2021-03-14
Add https://github.com/ashwin153/pre-commit

ashwin153:
- Support for the `vagrant-validate`  hook which runs `vagrant validate` on Vagrantfiles.
- Useful for linting changes to development environment configurations.


all-repos.yaml
```diff
@@ -176,3 +176,4 @@
 - https://github.com/cpplint/cpplint
 - https://github.com/MarcoGorelli/abs-imports
 - https://github.com/domdfcoding/flake2lint
+- https://github.com/ashwin153/pre-commit
```
asottile, line 179: ideally this repository wouldn't be named "pre-commit" or "pre-commit-hooks" as those are easily confusable with the framework itself

it seems this repository is `system`-only hooks as well which might not be very useful for consumers (where they could just make a `local` hook of the same thing)

also you can simplify your `files` regex expression to just `files: Vagrantfile`

ashwin153: - Renamed the repository to `pre-commit-vagrant` to disambiguate. Lmk if there's a more appropriate name you have in mind.
- I think there's precedent for this; [hadolint] does a similar validation of Dockerfiles in a `language: system` pre-commit hook. But, I suppose that it is not `system`-only. Are `system` pre-commit hooks considered an anti-pattern?
- Simplified the `files` regex expression.
- 
[hadolint]:
  https://github.com/hadolint/hadolint/blob/master/.pre-commit-hooks.yaml

asottile: `system` is kind of the escape hatch for hooks -- since pre-commit can't manage the installation for you

asottile (approved):

PR #476, opened 2021-03-15
Add dotnet-format to the list of hooks

chriselion:
Added in https://github.com/dotnet/format/pull/872


rkm:
Might be worth holding on this until they tag a new release with the added p-c-config

asottile:
oops

PR #483, opened 2021-03-23
Add oelint-adv

noahp:
Linter for bitbake (yocto) recipe files.


asottile (approved):
it might be cool to add the yocto file extensions to https://github.com/pre-commit/identify -- then you can utilize `types: [...]` instead of needing a `files: ` regex -- just a thought!

PR #485, opened 2021-04-05
Add pre-commit-kustomize hook

sharkymcdongles:
This hook allows you to run kustomize against repos to detect errors.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -182,3 +182,4 @@
 - https://github.com/AleksaC/mirrors-cfn-nag
 - https://github.com/cmake-lint/cmake-lint
 - https://github.com/priv-kweihmann/oelint-adv
+- https://github.com/sharkymcdongles/pre-commit-kustomize


all-repos.yaml
```diff
@@ -182,3 +182,4 @@
 - https://github.com/AleksaC/mirrors-cfn-nag
 - https://github.com/cmake-lint/cmake-lint
 - https://github.com/priv-kweihmann/oelint-adv
+- https://github.com/sharkymcdongles/pre-commit-kustomize
```
asottile, line 185: seems this tool is written in `go` -- did you see if upstream would be willing to add the metadata directly?  that would avoid the difficulty of installing, etc.

sharkymcdongles: I just wrote the hook for personal use and figured I'd pay it forward in case someone else had the same use case. What metadata would need to be added? Happy to submit a PR to them to add it. I checked here: https://pre-commit.com/index.html#golang

But it seems to just expect the go source code and says nothing about metadata needing to be added. If it just needs something installable via go get then it should already be able to do that. I can test it if this is what you mean.

asottile: similar to how you added a `.pre-commit-hooks.yaml`, one could be added to the original repository (with `language: golang`)

sharkymcdongles: I will submit to them as a PR and see.

PR #488, opened 2021-04-09
Add travis-lint

proinsias:
Adds a linter of `.travis.yml` files.

This is taken from travis-ci/travis.rb#307 which has been left to languish on the vine by the `travis.rb` maintainers.


asottile:
would prefer this from the source repository -- also this will not be properly versionable I don't think (?)

honestly travis will probably disappear into irrelevance soon, I don't really know of anyone that's still using them in earnest

proinsias:
No problem – was just playing around with creating new hooks!

PR #490, opened 2021-04-14
Add pre-commit hooks for android-kotlin

jggomez:
Add pre-commit hooks for android-kotlin

https://github.com/jggomez/pre-commit-android-kotlin
Screen Shot 2021-04-13 at 23 15 13

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -182,3 +182,4 @@
 - https://github.com/AleksaC/mirrors-cfn-nag
 - https://github.com/cmake-lint/cmake-lint
 - https://github.com/priv-kweihmann/oelint-adv
+- https://github.com/jggomez/pre-commit-android-kotlin


asottile:
your `unittests` hook can be replaced with:

```
language: system
entry: ./gradlew test
pass_filenames: false
always_run: true
```

actually, probably most of them can be adjusted to that as well?  as they're written right if you stage a lot of filenames you'll get multiple invocations of each (try with `--verbose`)

jggomez:
Ok, I have remove the script directory. Thanks

jggomez:
> your `unittests` hook can be replaced with:
> 
> ```
> language: system
> entry: ./gradlew test
> pass_filenames: false
> always_run: true
> ```
> 
> actually, probably most of them can be adjusted to that as well? as they're written right if you stage a lot of filenames you'll get multiple invocations of each (try with `--verbose`)

Hi, are there other any issues?

asottile (approved):

PR #491, opened 2021-04-14
Add pre-commit hook for Dansabel Ansible linter to all-repos.yml

jokjr:


README.md
```diff
@@ -5,3 +5,5 @@ pre-commit.com
 [![Build Status](https://github.com/pre-commit/pre-commit.com/workflows/deploy/badge.svg)](https://github.com/pre-commit/pre-commit.com/actions)
 
 This powers https://pre-commit.com
+
+To add a hook, add a link to your repository in `all-repos.yaml` and create a pull request.
```
asottile, line 9: please don't touch the README, thanks

jokjr:
@asottile Thanks, I took the README change out and force-pushed :-)

all-repos.yaml
```diff
@@ -182,3 +182,4 @@
 - https://github.com/AleksaC/mirrors-cfn-nag
 - https://github.com/cmake-lint/cmake-lint
 - https://github.com/priv-kweihmann/oelint-adv
+- https://github.com/semaphor-dk/dansabel
```
asottile, line 185: https://github.com/semaphor-dk/dansabel/blob/e2552c2fd5b076e161ee7fd81b06ebbc5fa9647b/.pre-commit-hooks.yaml#L10-L11

this script has 3rd party dependencies so it won't work out of the box -- you're best to use `language: python` and set up the setuptools metadata so your tool gets installed properly

jokjr: Thank you for your time and for the review. :-)

I agree with you in principle, but to be able to provide exact results it needs to use the exact same version of `ruamel.yaml` and `jinja2` as provided by the operating system - if the libraries diverge from the parsers used by Ansible (as installed with the OS package manager) the user ends up with false negatives/positives.
- The dependencies for that script (`jinjalint.py`) are already pulled in when you install `ansible`, so I thought it might be alright to sidestep automatic package management.
- The `pre-commit.sh` shell script also uses os-level dependencies, but it's written to serve as a freestanding pre-commit hook in its own right, so it might be less relevant for inclusion in `pre-commit`. Maybe the cleanest solution there would be to just remove `dansabel-all` from `.pre-commit-hook.yaml`.

I'll look into a potential solution for this; I totally understand that you might want all the "official" `pre-commit.com` repositories to be installable without hassle, but I'm not sure exactly how to cut the cake in this case.

asottile: > are already pulled in when you install `ansible`

this assumption isn't necessarily true, ansible could be installed in a virtualenv or to a different python

jokjr: That's a good point, which further complicates the issue. :-(

asottile: the general approach has been to provision an environment at the latest, and if the user needs to customize the dependencies they can do so via `additional_dependencies`

jokjr: I updated it and it seems to work. There was a problem with my `setup.py`: I had `requires=` instead of `install_requires=`, so the dependencies were not being installed. Now they are.

I also removed the shell script (from the pre-commit.com list of hooks) as I think there are existing hooks in this framework that do the same.

asottile (approved):

jokjr:
(Force-pushed a rebase on `master`)

jokjr:
Force-pushed again to fix merge conflict with #595.
@asottile is there anything left blocking a merge to master?

PR #498, opened 2021-04-21
Add jsonnetfmt hook

chrismgrayftsinc:


all-repos.yaml
```diff
@@ -184,3 +184,4 @@
 - https://github.com/priv-kweihmann/oelint-adv
 - https://github.com/jggomez/pre-commit-android-kotlin
 - https://github.com/Carreau/velin
+- https://github.com/chrismgrayftsinc/pre-commit-hooks
```
asottile, line 187: it would be preferable to name this repository something different from `pre-commit-hooks` such that it isn't easily confused with `pre-commit/pre-commit-hooks` -- maybe `jsonnetfmt` or something similar?  otherwise looks good

chrismgrayftsinc: Done

asottile (approved):

PR #500, opened 2021-04-25
Added pre-commit-fish

hugoh:
Added https://github.com/hugoh/pre-commit-fish to the list of hooks.


asottile:
I think for the `fish_syntax` hook you can replace it directly with:

```yaml
    entry: fish --no-execute
    language: system
```

hugoh:
> I think for the `fish_syntax` hook you can replace it directly with:
> 
> ```yaml
>     entry: fish --no-execute
>     language: system
> ```

Good point. Updated at https://github.com/hugoh/pre-commit-fish/pull/1

Thanks.

asottile (approved):

PR #503, opened 2021-04-29
Add optimize-png-hooks to list

redwarp:
This hook runs on png files, using oxipng to optimize pngs in a predictable manner.
Requires rust.


asottile (approved):
awesome!

PR #509, opened 2021-05-11
add pre-commit-license-checks to the list of available hooks

nbyl:
Here are two hooks to check the used licenses in a project against a dedicated allow list.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -188,3 +188,4 @@
 - https://github.com/zricethezav/gitleaks
 - https://github.com/hugoh/pre-commit-fish
 - https://github.com/redwarp/optimize-png-hooks
+- https://github.com/nbyl/pre-commit-license-checks


asottile:
- your console script entry here is incorrect: https://github.com/nbyl/pre-commit-license-checks/blob/97ebfa89dbeb65af70aeef537974a03ab4e726f2/setup.py#L16
- these should use executables instead of `python -m` for better portability (and so they can't leak into the working directory -- try `touch license_checks.py` and your current hooks will break: https://github.com/nbyl/pre-commit-license-checks/blob/97ebfa89dbeb65af70aeef537974a03ab4e726f2/.pre-commit-hooks.yaml#L5

nbyl:
@asottile Thank you for the review. I have applied your suggestions and released a new version which hopefully resolves the issues.

asottile (approved):

PR #528, opened 2021-06-22
Add pre-commit-xmllint

pkubanek:
Allows checks with output of xmllint --format. Reuquires xmlling (part of libxml) binary.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -193,3 +193,4 @@
 - https://github.com/dfm/black_nbconvert
 - https://github.com/crate-ci/typos
 - https://github.com/snakemake/snakefmt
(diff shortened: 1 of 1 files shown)


asottile:
this doesn't provide much, if any, value over:

```python
-   repo: local
    entry: xmllint --format
    types: [xml]
   ...
```

pkubanek:
Well the problem is that xmllint --format doesn't allow in-place reformat of the files (similar to e.g. black for Python). I understand it's easy to save --format output to tempfile and overwrite the original, but that will not get the required signaling if files changes etc,. (as --format return code is 0 when file can be parsed). A bit awkward, but we found the tool useful as we format XMLs with xmllint.

PR #534, opened 2021-07-14
Add the repository `detect-direct-checkins`

IngoMeyer441:
The `detect-direct-checkins` hook checks if configured branches contain
only merge commits (-> commits with two or more parents).

This check is similar to the official `no-commit-to-branch` hook but
allows the creation of merge commits. Fixes [issue 574 in the 
pre-commit-hooks repository](https://github.com/pre-commit/pre-commit-hooks/issues/574).

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -196,3 +196,4 @@
 - https://github.com/regebro/pyroma
 - https://github.com/tox-dev/tox-ini-fmt
 - https://github.com/janosh/format-ipy-cells
+- https://github.com/IngoMeyer441/detect-direct-checkins


asottile:
1. please don't use the name `pre-commit`, you are not part of the project
2. it is "pre-commit", not "Pre-commit"
3. you've set `stages: [post-commit]` which cannot influence the result of a commit so your thing doesn't work
4. merge commits don't trigger the `pre-commit` hook (only `pre-merge-commit`) so I don't think your thing is needed at all?
5. I believe this is a derivative work so it should abide by the [license of pre-commit/pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks/blob/master/LICENSE) (very easy permissive license, you just have to distribute a copy of it with your derivative work)

IngoMeyer441:
> please don't use the name pre-commit, you are not part of the project

I only used the name `pre-commit` to make it easier to identify that it is **for** the pre-commit project. Many other third party hooks on https://pre-commit.com/hooks.html also use the prefix `pre-commit-`.

> it is "pre-commit", not "Pre-commit"

Ok, will fix that.

> you've set stages: `[post-commit]` which cannot influence the result of a commit so your thing doesn't work

It must be a post-commit hook since my hook checks the created commit object itself. In the `pre-commit` stage it would be possible to make a direct check-in to a disallowed branch and the check would only fail on the next commit.

The README explains that the hook must be activated with

```bash
pre-commit install --hook-type post-commit
```

> merge commits don't trigger the `pre-commit hook` (only `pre-merge-commit`) so I don't think your thing is needed at all?

My hook does not need the `pre-commit` hook stage at all. Merge commits don't need a test because they are always allowed (in my checks).

> I believe this is a derivative work so it should abide by the license of pre-commit/pre-commit-hooks (very easy permissive license, you just have to distribute a copy of it with your derivative work)

I based its interface on the `no-commit-to-branch` hook, I am not sure if this is really a derivation. But since I also used the MIT license I can simply add a copyright notice to the `LICENSE` file.

asottile:
> > please don't use the name pre-commit, you are not part of the project
> 
> I only used the name `pre-commit` to make it easier to identify that it is **for** the pre-commit project. Many other third party hooks on https://pre-commit.com/hooks.html also use the prefix `pre-commit-`.
> 

yes, I am trying to be better about protecting the brand than I previously was.

> > you've set stages: `[post-commit]` which cannot influence the result of a commit so your thing doesn't work
> 
> It must be a post-commit hook since my hook checks the created commit object itself. In the `pre-commit` stage it would be possible to make a direct check-in to a disallowed branch and the check would only fail on the next commit.
> 
> The README explains that the hook must be activated with
> 
> ```shell
> pre-commit install --hook-type post-commit
> ```

`post-commit` is too late to influence whether something happens or not -- you can't prevent a commit after it already happens


> > merge commits don't trigger the `pre-commit hook` (only `pre-merge-commit`) so I don't think your thing is needed at all?
> 
> My hook does not need the `pre-commit` hook stage at all. Merge commits don't need a test because they are always allowed (in my checks).
> 
> The check is inteneded for branching models like the [gitflow](https://nvie.com/posts/a-successful-git-branching-model/) model. In this model some branches must only contain merge commits and no direct checkins. For this model, the official `no-commit-to-branch` is not sufficient because it also blocks merge commits. My hook allows to merge changes into a monitored branch but will fail as soon as any direct check-in (a commit with only one parent) is found.
> 

what I'm saying is, `no-commit-to-branch` **already allows merge commits because a merge commit does not trigger the `pre-commit` hook** so I posit that there's no need for your tool at all

IngoMeyer441:
> yes, I am trying to be better about protecting the brand than I previously was.

Ok, I'm going to rename the repository.

> what I'm saying is, `no-commit-to-branch` already allows merge commits because a merge commit does not trigger the pre-commit hook so I posit that there's no need for your tool at all

This is only partially correct. Let's consider a repository with a  `release` and a `develop` branch and the `release` branch must only contain merge commits (enforced by `no-commit-to-branch`):

```bash
git checkout develop
# Do some stuff
git commit
git checkout release
git merge --no-ff develop  # OK! pre-commit hook is not run
# Wait, there is typo in the commit message...
git commit --amend  # Fails! pre-commit hook is triggered and `no-commit-to-branch` prevents the commit creation

# Let's check if the release passes all pre-commit tests (for example in CI pipeline)
pre-commit run --all-files  # Fails! `no-commit-to-branch` detects we are on the `release` branch
```

Since `no-commit-to-branch` can be triggered for merge commits, it is broken and only works in some situations.

> `post-commit` is too late to influence whether something happens or not -- you can't prevent a commit after it already happens

Yes, you are right. Because of this I first tried to create a `pre-commit` stage hook to reliably detect if a merge commit is going to be created or not but I couldn't find a test which works in all situations. So in my opinion, it is better to get correct test results on a already created commit (so I know I must delete or rebase this commit) than getting a test result which could be a false positive.  Anyway, hooks run on local clones are never enforcing, a user could simply not install the pre-commit framework or skip all tests with the `--no-verify` flag. Enforcing checks must be run as part of a CI system. In a CI run my hook would detect a direct check-in and the CI pipeline would fail, so I don't see a real problem in using the `post-commit` stage.

Ok, the project description is misleading and I should rename the repository to something like `detect-direct-checkins` because it is only a detection and does not prevent wrong commits from happening.

asottile:
if you can't prevent it, I don't see the point.

you'd also just as easily `SKIP=no-commit-to-branch git commit --amend` (though I'd strongly argue you should *never* amend a merge commit making a dirty merge)

IngoMeyer441:
The point is, that `no-commit-to-branch` generates false positives and fails in CI pipelines for merge commits, so it is broken.
My proposed check behaves correctly in CI pipelines.

asottile:
> 
> 
> The point is, that `no-commit-to-branch` generates false positives and fails in CI pipelines for merge commits, so it is broken.
> My proposed check behaves correctly in CI pipelines.

only because yours doesn't run (`post-commit` is not triggered in CI) and can't fail (post-commit cannot exit nonzero) -- I don't see that as a solution

IngoMeyer441:
Of course, `pre-commit` must be executed like this:
```bash
pre-commit run --hook-type post-commit
```
It is describe in the [README](https://github.com/IngoMeyer441/detect-direct-checkins/blob/master/README.md) of the project.

asottile:
> 
> 
> Of course, `pre-commit` must be executed like this:
> 
> ```shell
> pre-commit run --hook-type post-commit
> ```
> 
> It is describe in the [README](https://github.com/IngoMeyer441/detect-direct-checkins/blob/master/README.md) of the project.

right and if you need two invocations and a custom parameter that's worse than using `SKIP` which already exists and requires no extra code

IngoMeyer441:
A check which is skipped in CI is useless and can be removed.

asottile:
I agree

PR #535, opened 2021-07-14
all-repos.yaml add format-ipy-cells hook

janosh:
Code formatter for cell delimiters in interactive iPython notebooks (`# %%`).

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -195,3 +195,4 @@
 - https://github.com/snakemake/snakefmt
 - https://github.com/regebro/pyroma
 - https://github.com/tox-dev/tox-ini-fmt
+- https://github.com/janosh/format-ipy-cells


asottile (approved):
I would recommend renaming your top level module from `hooks` to something unique so you don't collide when installed outside of pre-commit.  also you're using `find_namespace:` but you don't have namespace packages

janosh:
Thanks for the advice. Moved `format_ipy_cells.py` to root level and removed `find_namespace` in https://github.com/janosh/format-ipy-cells/commit/e073a0479617169b24a99d87bd8ea9f31ea02edd.

asottile:
you'll want `py_modules` if you do that btw

janosh:
Changed my mind afterwards. Splitting into `main.py` and `helpers.py` inside `format_ipy_cells/` made it easier to work with. Good to know about `py_modules` for the future though.

asottile:
you still want `find:` and not `find_namespace:` -- the latter is only for namespace packages: https://github.com/pre-commit/pre-commit/blob/fe436f1eb09dfdd67032b4f9f8dfa543fb99cf06/setup.cfg#L24

asottile:
oh I guess it is, but you  probably don't want it to be a namespace package?  you're not going to add another thing under `format_ipy_cells/` in a different package right?

janosh:
No but I also don't like to add `__init__.py` files. They offend my sensibilities. 😆

PR #538, opened 2021-07-23
Add hook for darker

arendondiosa:
https://github.com/akaihola/darker/#using-as-a-pre-commit-hook


asottile:
this seems incorrect, perhaps due to copy paste? https://github.com/akaihola/darker/blob/9006074ce450d416de3db9b5f5faa465c645695c/.pre-commit-hooks.yaml#L7

PR #541, opened 2021-08-03
add pre-commit-gitlab-ci

FalcoSuessgott:
add pre-commit-gitlab-ci

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -196,3 +196,4 @@
 - https://github.com/regebro/pyroma
 - https://github.com/tox-dev/tox-ini-fmt
 - https://github.com/janosh/format-ipy-cells
+- https://github.com/FalcoSuessgott/lint-gitlab-ci


asottile:
a few things:

- please don't use the `pre-commit` name for your project as you're not affiliated
- there's already at least 1 hook which does the same thing in the list already -- perhaps collaborate with that individual?

FalcoSuessgott:
> * please don't use the `pre-commit` name for your project as you're not affiliated

Understood. Updated the projectname to `lint-gitlab-ci`. :)

> there's already at least 1 hook which does the same thing in the list already -- perhaps collaborate with that individual?

Tried all of them they either they didnt work with the required Gitlab Access Token or they didnt supported merged gitlab-ci configurations. Also I wanted to have a project that does not depend on pip modules, my approach just uses `curl` and `jq`.

asottile:
> 
> 
> > * please don't use the `pre-commit` name for your project as you're not affiliated
> 
> Understood. Updated the projectname to `lint-gitlab-ci`. :)
> 
> > there's already at least 1 hook which does the same thing in the list already -- perhaps collaborate with that individual?
> 
> Tried all of them they either they didnt work with the required Gitlab Access Token or they didnt supported merged gitlab-ci configurations. Also I wanted to have a project that does not depend on pip modules, my approach just uses `curl` and `jq`.

curl and jq are going to be significantly less portable than using `urllib.request` + `json` from the pythn standard library -- maybe try with those ?

FalcoSuessgott:
> curl and jq are going to be significantly less portable than using `urllib.request` + `json` from the pythn standard library -- maybe try with those ?

I agree.. but I got a version now that works for me and I dont want to rewrite it ..it has been a while since I used python :D

Would appreciate it if you decide to merge the PR otherwise its ok for me :)

PR #543, opened 2021-08-10
Prepare frigate integration

jrottenberg:
https://github.com/rapidsai/frigate#pre-commit-hook

Frigate is a very nifty tool << for automatically generating documentation for your Helm charts.>>

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -197,3 +197,4 @@
 - https://github.com/tox-dev/tox-ini-fmt
 - https://github.com/janosh/format-ipy-cells
 - https://github.com/FalcoSuessgott/lint-gitlab-ci
+- https://github.com/rapidsai/frigate


asottile:
a few things:
- the readme should use `>> .pre-commit-config.yaml` probably so it doesn't clobber existing usages
- the hook is configured with `always_run: true` + `require_serial: true` -- this does not guarantee it only runs in one invocation which seems to be the intention.  I *think* you were trying to not pass filenames because it does not consume them in which case you should use `pass_filenames: false`

jrottenberg:
ok in progress, https://github.com/rapidsai/frigate/pull/36/files let me know if I'm missing anything.

jrottenberg:
@asottile , should be all good now.

asottile (approved):

jrottenberg:
thanks a lot, is there a way to add a quick description and set the hook type to 'helm' ?

asottile:
@jrottenberg that's all derived from the `description` and `types` fields in the hook metadata itself

jrottenberg:
gotcha, I'll update and wait for the next cron run, thanks  again !

jrottenberg:
I see yeah, I can reproduce locally. Not sure what to do from it, the hook classification is derived from https://github.com/pre-commit/identify which in no case will auto detect helm files... and for good reason.

I can remove the type, or patch your build script to allow a fallback if `identify` cannot detect the type and rely on the 'suggested' value (from the .pre-commit-hooks.yaml type)

asottile:
your configuration is unusable _with pre-commit itself_ -- `identify` `types` is how it does file matching

jrottenberg:
Ah yes of course I'm not using that in the pre-commit (run always regardless of the files changed), I wrongly assumed it was metadata for pre-commit.com.

I'll revert so and keep only the description...


Sorry about that.

PR #545, opened 2021-08-20
Add `xmllint` check

comkieffer:
Add pre-commit hook to validate xml using xmllint.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -198,4 +198,4 @@
 - https://github.com/janosh/format-ipy-cells
 - https://github.com/FalcoSuessgott/lint-gitlab-ci
 - https://github.com/rapidsai/frigate
-- https://github.com/comkieffer/pre-commit-xmllint
+- https://github.com/comkieffer/xml-linter-hook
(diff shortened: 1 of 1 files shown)


asottile:
dupe of #528

comkieffer:
I'd push to re-open this. Unlike the linked MR, this one installs `xmllint` into a [docker container](https://github.com/comkieffer/pre-commit-xmllint/blob/master/Dockerfile). The linked MR depended on `xmllint` being installed locally AFAIK.

asottile:
ok then:

- please do not name your thing `pre-commit` as it is not affiliated with the project
- it is "pre-commit", not "Pre-commit" or "Pre-Commit"
- the default should probably have `args: [--format]` ?

comkieffer:
1) Naming: I don't have a problem renaming the repo. The `pre-commit-{tool}` naming scheme seems quite common in the existing set of hooks. Are there any guidelines on acceptable names somewhere? 

2) `--format` is not an in-place operation, the result is sent to stdout. Initially, I toyed with a script to wrap `xmllint` to get that behaviour, but couldn't find a simple way to get the name of the file being processed. The linked MR side-steps the issue by not passing any user-supplied `args` to `xmllint` which breaks validation (which needs `--schema schema_url` for example).

asottile:
> The pre-commit-{tool} naming scheme seems quite common in the existing set of hooks.

indeed, I'm trying to be better now about protecting branding than previously

comkieffer:
I've changed the repo url to: https://github.com/comkieffer/xml-linter-hook 

Everything should be good to go.

asottile (approved):

PR #546, opened 2021-08-20
Clarify naming rules for hook repositories

comkieffer:
Update the "Supported Hooks" page to clarify that hook repositories should avoid the `pre-commit` name.


asottile:
this shouldn't be present on the public website, perhaps a contributing guide but not in user-facing content

PR #548, opened 2021-08-25
Add missing argument mention

revolter:


asottile:
this whole line should probably just be removed or rewritten to generalize that "there may be more options needed when using specific hook types" as it was written in a time where there was only one hook type with special arguments

revolter:
Personally, I find it clear enough that it's not a complete instruction, but it took me some time to figure out that I need the `--hook-stage` argument, especially for the case mentioned in this instruction (`commit-msg`). So I think that this change, at least improves the current state.

PR #549, opened 2021-08-29
Add jekyll-relative-url-hook

klieret:
I've created a hook for Jekyll websites that checks whether absolute links `[absolute link](/absolute/paths)` are used, rather than `[absolute link]({{site.baseurl}}/relative/path)` or similar constructs. More information is in the readme.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -199,3 +199,4 @@
 - https://github.com/FalcoSuessgott/lint-gitlab-ci
 - https://github.com/comkieffer/xml-linter-hook
 - https://github.com/jackdewinter/pymarkdown
+- https://github.com/klieret/jekyll-relative-url-check


all-repos.yaml
```diff
@@ -198,3 +198,4 @@
 - https://github.com/janosh/format-ipy-cells
 - https://github.com/FalcoSuessgott/lint-gitlab-ci
 - https://github.com/comkieffer/xml-linter-hook
+- https://github.com/klieret/jekyll-relative-url-hook/
```
asottile, line 201: - none of the other urls have a trailing slash
- hook ids should be `dashed-names`
- entrypoints should be `dashed-names`
-

klieret:
Hi @asottile  Could you review this PR again? I followed all of your suggestions

asottile (approved):

PR #552, opened 2021-09-03
docs: add text-prepender

dontirun:


asottile:
I believe there's an existing tool which already does this: https://github.com/Lucas-C/pre-commit-hooks#insert-license

dontirun:
I didn't initially see that hook. They both are similar, but slightly different. 

1. The existing tool you specify a file type and a comment style 
2. The tool that I created it goes over a list of supported file types and edits all of the supported types 

That being said, if the tools are similar enough I'll close this pull request

asottile:
where possible I think it makes sense to collaborate 👍

PR #555, opened 2021-09-13
Add cspell to all_repos.yaml

janosh:
[A spell checker for code](https://github.com/streetsidesoftware/cspell).


asottile:
this should probably be `types: [text]` otherwise this is going to try and spellcheck binary files which doesn't really make much sense: https://github.com/streetsidesoftware/cspell-cli/blob/1a3acfb5573804594b1421cc8e67662fedaa5c32/.pre-commit-hooks.yaml#L10

asottile (approved):

PR #557, opened 2021-09-16
Fix gitlab-ci cache example

bsecker:
Add a trailing slash to the `PRE_COMMIT_HOME` variable so that gitlab-ci caches correctly


asottile:
the gitlab docs seem to indicate that no slash is necessary -- I suspect your issue is elsewhere?

asottile:
yeah this isn't correct -- I'm using this sucessfully here: https://github.com/pre-commit-ci-demo/demo/blob/cf38b7f21be6050bbf9934c59657174d60c38f7b/.gitlab-ci.yml#L4-L10

PR #559, opened 2021-09-25
helm docs generator

jrottenberg:
Following up from https://github.com/pre-commit/pre-commit.com/pull/543

At this point frigate and helm-docs are stable, and should not break the build.


all-repos.yaml
```diff
@@ -200,3 +200,5 @@
 - https://github.com/comkieffer/xml-linter-hook
 - https://github.com/jackdewinter/pymarkdown
 - https://github.com/klieret/jekyll-relative-url-check
+- https://github.com/rapidsai/frigate
+- https://github.com/norwoodj/helm-docs
```
asottile, line 204: shouldn't `helm-docs` be `language: go` instead of `language: script` ?

jrottenberg: Yes, If you want to install helm-docs itself, the intent most likely is to rely on the user to have the binary installed.

Most likely the user doesn't have the golang build chain, hence the `script`

jrottenberg: @asottile https://github.com/antonbabenko/pre-commit-terraform/blob/master/.pre-commit-hooks.yaml#L9 has a similar setup : expose a hook that relies on the user to install the binary before execution, the hooks just making sure the tool has run before commmit.

jrottenberg: Fair enough , I'll open a PR on helm-docs to add a new target that uses `golang` instead of `script`, so use could choose. Would that work ?

jrottenberg: ok https://github.com/norwoodj/helm-docs/pull/114 has been merged - good recommandation, it might take a while to go through existing hooks, but I do agree it makes the setup for users or ci, much simpler.

asottile: sweet!  this patch just needs a quick rebase and I'm happy to merge this!

asottile (approved):

asottile (approved):

jrottenberg:
Thanks !

PR #564, opened 2021-09-28
chore: add clang-format-precommit

henryiii:
Uses 1-2 MB PyPI package on all platforms.

Let me know if there's any special ordering you'd like, just stuck it at the end.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -200,3 +200,4 @@
 - https://github.com/comkieffer/xml-linter-hook
 - https://github.com/jackdewinter/pymarkdown
 - https://github.com/klieret/jekyll-relative-url-check
+- https://github.com/ssciwr/clang-format-precommit


asottile:
at the end is fine -- I would prefer it it were named `clang-format-hook` instead  -- don't want to confuse it with `pre-commit` itself

henryiii:
But `-hook` doesn't really make it clear what it's used for (it's pre-commit specific). I'll check and see if they are interested in moving it. I think it still will work for early adopters if renamed.

henryiii:
I probably inspired the current name by pointing at cmake-format-precommit as an example: https://github.com/ssciwr/clang-format-wheel/issues/8#issue-1009777067

asottile:
alternatively -- if you want to add `types_or` support to https://github.com/pre-commit/pre-commit-mirror-maker then I could create a `mirrors-clang-format` and then we wouldn't need any external management

henryiii:
x-ref: https://github.com/pre-commit/pre-commit-mirror-maker/issues/69

asottile:
thanks for the work on `types_or` -- I set this up using it!  https://github.com/pre-commit/mirrors-clang-format

dokempf:
Hey, @asottile thanks for setting up this mirror. I prefer this option over the current solution of having a separate repository in the same namespace as the packaging repo (https://github.com/ssciwr/clang-format-hook). I would prefer to have the possibility to write a custom (less generic) Readme though. Would you accept PRs to this mirror repo or consider adding me to the repo so that I can edit the Readme? (I maintain the packaging repo https://github.com/ssciwr/clang-format-wheel)

asottile:
yeah feel free to send PRs -- though usually I try and keep the readmes as simple as possible and direct issues to the parent repositories (unless of course it's a bug in the configuration)

PR #577, opened 2021-11-01
Added add-msg-issue-suffix-hook

rnjesuz:
pre-commit hook that suffixes Jira issues to the commit message

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -206,3 +206,4 @@
 - https://github.com/streetsidesoftware/cspell-cli
 - https://github.com/rapidsai/frigate
 - https://github.com/norwoodj/helm-docs
+- https://github.com/rnjesuz/add-msg-issue-suffix-hook


asottile:
there's already at least two hooks which accomplish this already in the hooks html -- consider contributing to those!

rnjesuz:
> 
> 
> there's already at least two hooks which accomplish this already in the hooks html -- consider contributing to those!

Can you point me to them? 
The only one I'm seeing is https://github.com/avilaton/add-msg-issue-prefix-hook, which prefixes the jira issue. Mine suffixes it.
But mine also does a little more. As of last release it allows for optionally appending the issue's title as well as it's name

asottile:
> 
> 
> > there's already at least two hooks which accomplish this already in the hooks html -- consider contributing to those!
> 
> Can you point me to them? The only one I'm seeing is https://github.com/avilaton/add-msg-issue-prefix-hook, which prefixes the jira issue. Mine suffixes it. But mine also does a little more. As of last release it allows for optionally appending the issue's title as well as it's name

seems like your extensions would be an easy PR to that  one!

rnjesuz:
> 
> 
> seems like your extensions would be an easy PR to that one!

Doesn't seem trivial to me, especially with my added feature. It would need some refactoring and conditional checking.
I'm content with the state of the repo. 
You can deny the PR if you still don't think it's adding anything new.

asottile:
yeah I would prefer collaboration rather than fork/competition

PR #580, opened 2021-11-15
Add adamchainz/pre-commit-dprint mirror

adamchainz:


asottile (approved):
will need to resolve the conflict first -- but otherwise looks good

PR #581, opened 2021-11-16
Set Git merge=union for all-repos.yaml

adamchainz:
This should prevent conflicts from parallel PR’s like in #580. We’ve done this for [Django’s release notes](https://github.com/django/django/blob/12fe3224f5086161462faf614cad91f3fad32e78/.gitattributes#L7) for a while.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+all-repos.yaml merge=union


asottile (approved):
seems to work and order correctly -- we'll try this out and see how it goes

asottile:
github doesn't seem to use this file when resolving merges so I don't think this provides any benefit

adamchainz:
aw sad :(

PR #592, opened 2021-11-25
Add gersemi CMake formatter

BlankSpruce:
Anticipating question "Why `require_serial: true`?" here's rationale: gersemi already spawns additional processes to do the formatting faster in case of multiple files.


asottile (approved):
yep -- I know @DanielChabrowski from my twitch stream :)

PR #595, opened 2021-12-03
Add check-json5 to all-repos.yaml

maresb:
This is just like `check-json` from the official hooks, but it tries to load a file with [`json5`](https://pypi.org/project/json5/) instead of the builtin [`json`](https://docs.python.org/3/library/json.html). The [JSON5 spec](https://json5.org/) allows for comments, trailing commas, and more.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -205,4 +205,4 @@
 - https://github.com/scop/pre-commit-shfmt
 - https://github.com/BlankSpruce/gersemi
 - https://github.com/realm/SwiftLint
-- https://gitlab.com/bmares/pre-commit-check-json5
+- https://gitlab.com/bmares/check-json5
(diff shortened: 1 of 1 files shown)


all-repos.yaml
```diff
@@ -205,3 +205,4 @@
 - https://github.com/scop/pre-commit-shfmt
 - https://github.com/BlankSpruce/gersemi
 - https://github.com/realm/SwiftLint
+- https://gitlab.com/bmares/pre-commit-check-json5
```
asottile, line 208: nice!  would it be possible to rename it to just `check-json5` -- that would better differentiate it from being a "pre-commit" affiliated project vs just a project that can be integrated with pre-commit

asottile: additionally it would be best to avoid using the `pre_commit_hooks` namespace

maresb: Ah, ok... just a few minutes, I'll have to update the repos and the README. Thanks for the quick response! ...

asottile (approved):

all-repos.yaml
maresb: @asottile, done! And thanks in general for your awesome work. I'm a big fan. :smile:

PR #602, opened 2021-12-09
Add cirlceci orb validation to all-repos

bjd2385:
Adds a circleci orb pack and validation hook I've just written to the all-repos list.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -209,3 +209,4 @@
 - https://github.com/semaphor-dk/dansabel
 - https://github.com/adamchainz/pre-commit-oxipng
 - https://github.com/gitguardian/gg-shield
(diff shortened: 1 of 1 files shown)


asottile:
hi!  appreciate the tool, however since it is `system` / `script` it's not super useful out-of-the-box.  part of the benefit of pre-commit is to have the tool install the necessary dependencies so the user doesn't need to manage a bunch of things

additionally, the hardcoded `args: [src]` probably isn't suitable for all consumers

additionally, `require_serial: false` is the default so that's not needed

you probably also want `always_run: true` since it has `pass_filenames: false`

bjd2385:
Also, it's totally fine if this isn't accepted to the pre-commit.com registry/list, I just want to make sure it's correct and as useful to consumers as possible.

PR #611, opened 2022-01-06
Add apeschar/pre-commit-phan to all-repos.yaml

apeschar:
This PR adds apeschar/pre-commit-phan to the list of hooks, so that users can easily add [Phan](https://github.com/phan/phan) to their pre-commit checks.


asottile:
hi!  thanks for the PR, however:

- since this is implemented using `script` / `system` it isn't particularly useful out of the box for pre-commit users and doesn't provide much if any benefit over a `local` hook
- the repository is not affiliated with pre-commit so it should not use `pre-commit-` as a name

PR #615, opened 2022-01-13
Add databooks hook

murilo-cunha:


all-repos.yaml
```diff
@@ -210,3 +210,4 @@
 - https://github.com/adamchainz/pre-commit-oxipng
 - https://github.com/gitguardian/gg-shield
 - https://github.com/JohnnyMorganz/StyLua
+- https://github.com/datarootsio/databooks
```
asottile, line 213: I believe this incorrectly sets [`require_serial: true`](https://github.com/datarootsio/databooks/blob/a7b627b57a53d2562553e099684f5b32a3a6acc1/.pre-commit-hooks.yaml#L8)

murilo-cunha: Hmm why would this be incorrect? I thought it'd be best to `require_serial` as it'd write files to disk (similar to `black`)

asottile: that's not why black has `require_serial` -- black has `require_serial` because it does work using multiple processes

murilo-cunha: I see I see, [I've removed it](https://github.com/datarootsio/databooks/blob/main/.pre-commit-hooks.yaml)! Thanks for the explanation
And nice tool btw 🙃

asottile: thanks! you'll need to rebase this unfortunately due to conflicts

murilo-cunha: Done ✅

PR #616, opened 2022-01-14
Add check-conventional-commits

attakei:


asottile:
there's a few other commit-msg linters already in the list -- and some that support conventional commits -- it's probably better to collaborate with them?

rlindsgaard:
I think this looks awesome, I am _so_ adopting conventional commits!

Are there anyone but https://github.com/avilaton/add-msg-issue-prefix-hook? I couldn't find any linters in there.

In fact I just created https://github.com/rlindsgaard/pre-commit-commit-msg-hooks as I wanted to get acquainted with pre-commit and thought it would be an addition to the echo-system (I came in here to make a PR adding it). But maybe there's already something there I just couldn't find?

asottile:
for example, here's two much more battle tested commit message linters:
- https://github.com/commitizen-tools/commitizen
- https://github.com/jorisroovers/gitlint

PR #619, opened 2022-01-15
Adding sign-off hook

dynobo:


asottile:
while I kinda like the idea, I think this violates the whole idea of DCO since the committer is no longer _consenting_ to their signoff

dynobo:
Mh, tbh, I have no idea how this stands legally, e.g. if I could consent on signing-off all my commits _per se_ (by enabling this hook) or if it is required to do this more actively do this one by one through passing a cli-switch to `git`. 
AFAIK the signature alone is also of limited legal value, that's why in some project's you are still required to sign a special document digitally, which then gets filed.

What I know from _practice_ is, that teams/people are already automatizing this by e.g. smart aliases or scripts etc, because sometimes it is really a challenge to know, in which repos you have to sign and which you have not to. That's the practical idea behind it.

But it is a interesting question, would be cool to hear other opinions about it :-)

Another idea: I could also rebrand the hook to be more generic by calling it `commit-msg-append` and adjusting the examples/defaults. I'm quite sure the hook can be useful for other use-cases, too, e.g. when you want to add your department id to every message or something.

PR #620, opened 2022-01-16
Add byexample hook

eldipa:
`byexample` runs the code snippets in the documentation and validate them
as regression tests. From 10.4.0 it can be used as a `pre-commit` hook.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -211,3 +211,4 @@
 - https://github.com/JohnnyMorganz/StyLua
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
+- https://github.com/byexamples/byexample


all-repos.yaml
```diff
@@ -211,3 +211,4 @@
 - https://github.com/JohnnyMorganz/StyLua
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
+- https://github.com/byexamples/byexample
```
asottile, line 214: I don't think either of these options make sense: https://github.com/byexamples/byexample/blob/316db5b475a58bd96e7eb54510ab6929b8c88b70/.pre-commit-hooks.yaml#L6-L7

- `pass_filenames: true` -- this is the default
- `require_serial: true` -- `byexample` doesn't use multiprocessing as far as I can tell

I'm not sure `byexample` really makes sense as a pre-commit hook, the types of execution sound like they'll be too slow to make sense there and ~generally we discourage using test runners in hooks (the user's dependencies won't be available and they're too slow)

eldipa: Hello, thanks for the quick response.

> `pass_filenames: true` -- this is the default

Correct. It is kind of personal preference to make explicit the default parameters in the case that in a far future they change. I can remove it if it makes the hook config cleaner.

> `require_serial: true` -- `byexample` doesn't use multiprocessing as far as I can tell

Yes it does (and it took me a while to implementing it!). The option `--jobs=N` makes `byexample` to process N tests in parallel. It is not the default however.

> [...] sound like they'll be too slow [...]

Indeed, `byexample` adds an extra overhead but if it is too slow or not depends more of how many tests are scheduled and how much time they take. Because `pre-commit` passes the names of the files that changed only, the time would be roughly proportional to the size of that set.

But I guess this is true for any other hook that runs unit tests: how fast they will run will depend of how fast is each test.

Running also unit tests over a file that changed does not guarantee that you will catch all the bugs. I'm think in the case where `foo.py` depends on `bar.py` and a change in `bar.py` does not break any test in `bar.py` but it does in `foo.py`.

So I'm not a fan of running unit test in hooks either. Running them before each commit is too much but running them before a push or even manually makes more sense [reference](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages)

BTW, when you said "the user's dependencies won't be available". What did you mean? I didn't catch that part.

Thanks for your time!
Martin.

asottile: tools are installed in isolated environments so they won't have access to user dependencies (pre-commit never installs from the repository under test)

eldipa: I didn't know that! But giving it a second thought, it makes totally sense. 

Indeed, that is a real problem for any hook that loads and execute the user's code, hooks for unit-testing are particular effected.

Well, I cannot beat that! Merging the PR or not is up to you. From my side I will add a note in the documentation of `byexample` explaining all of this and setting expectations.

Thanks for all!

eldipa: Ah! jaj, sorry, I couldn't resist! I found a hack to "escape" from the environment set by `pre-commit` just before running the tests with `byexample`.

The trick is to have an auxiliary script `envpython.sh`

```shell
#!/bin/bash
source .project-environment/bin/activate
python $@
deactivate
```

Then, in the hook configuration add the `-x-shebang` argument as:

```yaml
repos:
-   repo: https://github.com/byexamples/byexample
    rev: 10.4.2
    hooks:
    -   id: byexample
        types_or: [markdown, python]
        args: [--jobs=4, -l=python, -l=shell, -x-shebang,  "python:%e ./envpython.sh %a" ]
        require_serial: false
```

With that, `byexample` will execute `envpython.sh` as the interpreter to run and test the Python examples, which in turn temporally activate the project's environment.

A hack, I know. Bu I really liked the outcome of this discussion.

asottile: neat -- thanks again!

PR #626, opened 2022-01-27
Disable smooth scrolling

chriskuehl:
I went to the pre-commit docs to try to find the details of the `fail` language and noticed that searching the page uses a smooth scroll animation which is pretty slow and a bit nausea-inducing:

https://user-images.githubusercontent.com/665269/151456067-3caad462-691d-4c4a-87c8-d44ba207ddc7.mov

I found a fix here which keeps the smooth scrolling when navigating between sections but not when using Ctrl-F to search: https://schepp.dev/posts/smooth-scrolling-and-page-search/

The CSS is a little bit complex, I'm happy to delete this PR if you don't care enough (or just turn off all smooth scrolling if you prefer).

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -23,3 +23,5 @@ $code-color: #e83e8c !default;
 
 $lead-font-size: 1.75rem !default;
 $lead-font-weight: 400 !default;
+
+$enable-smooth-scroll: false;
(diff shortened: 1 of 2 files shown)


scss/_variables.scss
```diff
@@ -23,3 +23,5 @@ $code-color: #e83e8c !default;
 
 $lead-font-size: 1.75rem !default;
 $lead-font-weight: 400 !default;
+
+$enable-smooth-scroll: false;
```
asottile, line 27: I think it would probably be fine to just turn it off entirely -- I don't think it really adds anything

asottile (approved):

PR #629, opened 2022-02-03
Add pre-commit-assert-changelog to all-repos

SamStudio8:


all-repos.yaml
```diff
@@ -213,3 +213,4 @@
 - https://github.com/JohnnyMorganz/StyLua
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
+- https://github.com/SamStudio8/pre-commit-assert-changelog
```
asottile, line 216: - there's no readme
- there's no license
- you've got a few security bugs waiting to happen (`shell=True`, `sys.path.append('.')`)
- executables are typically `kebab-case` not `snake_case`
- your project should not contain `pre-commit-` as it is not affiliated with pre-commit

SamStudio8: Thank you for taking a look!

* README and license added
* Removed shell=True and use sys.path.append("") as per expectation for when script dir is not available
* Executable is now `assert-changelog`
* README clarifies this is a `pre-commit` hook and not a `pre-commit` project but I am happy to change the repo -- I just followed what seemed to be the general consensus on github

asottile: pre-commit doesn't run with unstaged changes so I don't think this hook will work at all

SamStudio8: The behaviour I'm seeing is `pre-commit` emits a warning on unstaged changes and stashes them.  The list of unstaged files is inferred by `assert-changelog` (https://github.com/SamStudio8/pre-commit-assert-changelog/blob/main/assert_changelog/main.py#L108). It demonstrably [works for me](https://twitter.com/samstudio8/status/1489254738247835650) but happy to admit this is a very alpha project that perhaps should not be in your list.

PR #632, opened 2022-02-15
Black is no longer a beta product

pierreluctg:
Black is no longer a beta product, setting black rev to 22.1.0.

https://github.com/psf/black/releases/tag/22.1.0


asottile:
thanks for the PR but this is not meant to be kept up to date, just used as an example

pierreluctg:
> thanks for the PR but this is not meant to be kept up to date, just used as an example

@asottile I do understand that this is only an example, But I do not believe using a beta version of a tool is the best "example".

PR #633, opened 2022-02-24
Add a link to the site source

hexagonrecursion:
This should make it easier to find the source for those who want to contribute to the documentation


asottile:
hi!  thanks for the contribution but adding this as a nav item isn't correct (it breaks the nav system) and I'd rather not either way

hexagonrecursion:
> it breaks the nav system

What if I shorten the link text to "Edit"?

asottile:
it still breaks the nav system since there's no associated section

PR #634, opened 2022-02-24
Recommend `pipx run`

hexagonrecursion:
I find `pipx run` to be a much more convenient method of installing CLI tools than manually managing virtual environments. This is especially true for projects that are not written in Python


asottile:
please search the tracker next time https://github.com/pre-commit/pre-commit.com/pull/421

PR #635, opened 2022-02-24
Add REUSE tool

hexagonrecursion:
Add a new hook to the list: this one verifies the presence machine-readable license information.

Note: I am linking to fsfe.org instead of GitHub because https://reuse.software/dev/#tool links to https://git.fsfe.org/reuse/tool (click on the "REUSE helper tool" link)


asottile:
the maintainers of REUSE had a fairly toxic interaction a few years back so I don't really want to advertise their tool

thanks for understanding

PR #638, opened 2022-03-01
A linter for Sceptre stack configs

zaro0508:
Add a reference to the Sceptre pre-commit linter.  This linter checks
sceptre stack configurations[1]

[1] https://docs.sceptre-project.org/dev/docs/stack_config.html

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/Sceptre/sceptrelint


all-repos.yaml
```diff
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/Sceptre/sceptrelint
```
asottile, line 217: https://github.com/Sceptre/sceptrelint/blob/aee222f5565cb91748b8af23f09a0916600888e1/.pre-commit-hooks.yaml#L5-L6

- `pass_filenames: true` is the default
- `always_run: true` generally should not be set -- pre-commit will pass filenames for changed things to the tool so it runs selectively (otherwise when no files are changed the tool will receive 0 filenames and _probably_ do the wrong thing)

zaro0508: thanks for the tip @asottile .  Those configs have been removed.

asottile (approved):

PR #639, opened 2022-03-02
Add NOCOMMIT pre-commit hook

guillemglez:


all-repos.yaml
```diff
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/guillemglez/pre-commit-nocommit
```
asottile, line 217: there's no need for a whole repo for this, it can be done without code using [`pygrep`](https://pre-commit.com/#pygrep)

```yaml
-   repo: local
    hooks:
    - id: NOCOMMIT
      name: NOCOMMIT
      entry: \bNOCOMMIT\b
      language: pygrep
      types: [text]
```

guillemglez: Well thanks for the tip! I (wasn't that smart and) could not find anything in the website, so ended up making that repo... Perhaps I won't be the only one this helps...?

PR #640, opened 2022-03-04
Add yamlfix hook

wwuck:


all-repos.yaml
```diff
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/lyz-code/yamlfix
```
asottile, line 217: - this is probably incorrectly copy pasted from somewhere https://github.com/lyz-code/yamlfix/blob/f693d069eaf5f7afb7458c8206059e1ab6925df4/.pre-commit-hooks.yaml#L5
- this is redundant, if it is `yaml` it is already `file`: https://github.com/lyz-code/yamlfix/blob/f693d069eaf5f7afb7458c8206059e1ab6925df4/.pre-commit-hooks.yaml#L8
- this is the default: https://github.com/lyz-code/yamlfix/blob/f693d069eaf5f7afb7458c8206059e1ab6925df4/.pre-commit-hooks.yaml#L9
- this version doesn't add any specific features that this hook is using? https://github.com/lyz-code/yamlfix/blob/f693d069eaf5f7afb7458c8206059e1ab6925df4/.pre-commit-hooks.yaml#L10

wwuck: Thanks.  I'll try to submit a PR to that repo with fixes.

wwuck: Fixed in https://github.com/lyz-code/yamlfix/pull/168.

asottile (approved):

PR #641, opened 2022-03-06
Pipenv setup

jrottenberg:
Integration for https://github.com/Madoshakalaka/pipenv-setup

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/Madoshakalaka/pipenv-setup
--- a/install-local.py
+++ b/install-local.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import contextlib
+import distutils.spawn
 import hashlib
 import io
 import os.path
@@ -11,8 +12,6 @@ import shutil
 import subprocess
 import sys
 import tarfile
-
-import distutils.spawn
 
 
 if str is bytes:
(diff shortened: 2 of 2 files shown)


jrottenberg:
If you are ok with the first commit, I can squash them, i'll remove otherwise.

all-repos.yaml
```diff
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/Madoshakalaka/pipenv-setup
```
asottile, line 217: a few pieces of feedback:
- it's spelled "pre-commit" not "Pre-commit"
- there's no need for ".git" on the end of the `repo` url
- if `--pipfile` is always needed for the tool, perhaps it should be `entry: pipenv-setup sync --pipfile` ? https://github.com/Madoshakalaka/pipenv-setup/blob/22d04cb900adf2a6fcbaa7bd0dce9f4ea436ba80/.pre-commit-hooks.yaml#L4-L5
- this does nothing -- well actually it *could* do something in that it'll break if `setup.py` is `chmod +x` which you probably don't want: https://github.com/Madoshakalaka/pipenv-setup/blob/22d04cb900adf2a6fcbaa7bd0dce9f4ea436ba80/.pre-commit-hooks.yaml#L7

jrottenberg: 1. Easy fix
2. Where is the extra `.git` ?
3. `--pipfile` is a safe default, but use can overwrite it, if needed
4. Good point, https://github.com/Madoshakalaka/pipenv-setup/blob/22d04cb900adf2a6fcbaa7bd0dce9f4ea436ba80/.pre-commit-hooks.yaml#L8 is enough, L7 can be safely deleted ?

PR #642, opened 2022-03-08
Add rm_unneeded_f_str

dannysepler:
Docs here: https://github.com/dannysepler/rm_unneeded_f_str

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -216,3 +216,4 @@
 - https://github.com/standard/standard
 - https://github.com/Sceptre/sceptrelint
 - https://github.com/lyz-code/yamlfix
+- https://github.com/dannysepler/rm_unneeded_f_str


all-repos.yaml
```diff
@@ -214,3 +214,4 @@
 - https://github.com/shssoichiro/oxipng
 - https://github.com/datarootsio/databooks
 - https://github.com/standard/standard
+- https://github.com/dannysepler/rm_unneeded_f_str
```
asottile, line 217: - typically ids use dashes rather than underscores: https://github.com/dannysepler/rm_unneeded_f_str/blob/fb3a30dbf8a7eaad60b76bf761972af08f9d6540/.pre-commit-hooks.yaml#L1
- this is wrong: https://github.com/dannysepler/rm_unneeded_f_str/blob/fb3a30dbf8a7eaad60b76bf761972af08f9d6540/.pre-commit-hooks.yaml#L7

asottile (commented):
is this needed given `pyflakes` (via `flake8`) has [fstring is missing placeholders](https://github.com/PyCQA/pyflakes/blob/84da8cdaad574df7e692dff06ab561acc63d521c/pyflakes/messages.py#L278-L279) ?

dannysepler:
oh, i made this package _because_ of that flake8 error (since `autopep8` and `autoflake` both don't address it)

all-repos.yaml
dannysepler: fixed, thanks!

asottile: looks like you might also have to fix `rf` and `fr` strings -- glancing at the logic those might be broken

dannysepler: good catch! fixed by https://github.com/dannysepler/rm_unneeded_f_str/pull/8

asottile (approved):

PR #645, opened 2022-03-18
Document issue with gitlab non-constant build location

bagerard:
Document issue with gitlab non-constant build location that clashes with pre-commit's cache that expects a constant location.

This relates to https://github.com/pre-commit/pre-commit/issues/1133#issuecomment-985053269

### DIFF
--- a/sections/advanced.md
+++ b/sections/advanced.md
@@ -743,7 +743,7 @@ my_job:
 ```
 
 pre-commit's cache requires to be served from a constant location between the different builds. This isn't the default when using k8s runners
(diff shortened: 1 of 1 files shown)


sections/advanced.md
```diff
@@ -742,6 +742,9 @@ my_job:
       - ${PRE_COMMIT_HOME}
 ```
 
+pre-commit's cache requires to be served from a constant location between the different builds. This isn't the default when using k8s runners
+on Gitlab. In case you face the error `InvalidManifestError`, set `builds_dir` to something static e.g `builds_dir = "/builds"` in your `[[runner]]` config
```
asottile, line 746: while I don't care personally, someone else will send me another PR to change this to "GitLab" so it's probably best to do this here first!

PR #649, opened 2022-03-27
Adding grantjenks/blue pre-commit-hook

ChrisCarini:
Adding grantjenks/blue pre-commit-hook.

**Repo:** https://github.com/grantjenks/blue
**`.pre-commit-hooks.yaml` File:** https://github.com/grantjenks/blue/blob/master/.pre-commit-hooks.yaml

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -60,6 +60,7 @@
 - https://github.com/noahsark769/xcodeproj-sort-pre-commit-hook
 - https://github.com/jorisroovers/gitlint
 - https://github.com/psf/black
+- https://github.com/grantjenks/blue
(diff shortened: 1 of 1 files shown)


asottile:
> we monkeypatch to change these decisions to our own liking.

I think this is a bit too fragile to suggest

ChrisCarini:
> I think this is a bit too fragile to suggest

Hi @asottile ! Thank you for taking the time to look at this PR, as well as the underlying tool!

Would you mind elaborating on why you think `blue`'s approach is too fragile to suggest / have listed on the pre-commit.com hooks page? The maintainers have `blue` pinned to a specific version of `black` so there should never be any surprises or fragility (of course, so long as `black` does not change the underlying code a given tag/version points to).

PR #666, opened 2022-05-26
Add clj-kondo to the repos list

joelhess:


all-repos.yaml
```diff
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/clj-kondo/clj-kondo
```
asottile, line 222: is there a package-manager version of this available? as it is right now it doesn't provide much benefit over a `repo: local` hook since it is a `language: system` hook -- it's preferable to have the tool bootstrapped by a language pre-commit supports

it might also be possible to add support for whatever package manager is used here if you're interested in contributing one -- there's a rough outline of directions here: https://github.com/pre-commit/pre-commit/blob/main/CONTRIBUTING.md#adding-support-for-a-new-hook-language

joelhess: It's available on homebrew if that's what you're asking... https://github.com/clj-kondo/clj-kondo/blob/master/doc/install.md#brew-macos-and-linux

asottile: the point of pre-commit is it manages the tool versioning, installation, and setup for you so that you don't need to instruct your contributors on how to set up a bunch of tools

joelhess: I'll look into that. It's more than I was aiming for, but I get what you mean.

nicorikken: It seems that some pre-commit support for clojure packages through `tools.deps.alpha` or Leiningen would be great in this case: https://github.com/clj-kondo/clj-kondo/blob/master/doc/jvm.md

PR #668, opened 2022-06-04
Explicit `verbose == false` by default defining hooks

mondeja:
Explicitly declare that `verbose` is `false` by default in "Creating new hooks" section.


asottile:
hooks should not use this setting except for debugging so I do not want to document it here

asottile:
but also the messaging already implies the default

asottile:
eh whatever it doesn't hurt

PR #671, opened 2022-06-10
Add REUSE tool repository

nicorikken:
The CLI tool of the https://reuse.software/ project already had a pre-commit hook for a while: https://github.com/fsfe/reuse-tool/blob/master/.pre-commit-hooks.yaml By adding it to the official list it will be more convenient for others to discover and adopt REUSE.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/fsfe/reuse-tool


asottile:
https://github.com/pre-commit/pre-commit.com/pull/635#issuecomment-1049997786

nicorikken:
Hmm, I don't know the back story in this, but I dislike the fact you shut the door this way. Recently the REUSE development has stepped up the effort and community interaction. I hope you are open to talking it over or forgiveness and to give it some reconsideration. I am just here trying to improve the overall developer experience, not to get involved in previous fights. Luckily pre-commit is open-ended so people can still use the repository regardless. I will add the documentation at REUSE side because I still think pre-commit is a great solution that will be a help to many. Thanks for developing it!

PR #675, opened 2022-06-22
Add mz-lictools to the repos list

emzeat:
Provides pre-commit hooks to automagically manage the
license headers on source files.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/emzeat/mz-lictools


all-repos.yaml
```diff
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/emzeat/mz-lictools
```
asottile, line 222: there's already a battle-tested version of this tool listed in https://pre-commit.com/hooks.html https://github.com/Lucas-C/pre-commit-hooks

asottile: also not toxic GPL licensed

emzeat: Not sure why AGPL would matter or be toxic for this as using a copy-left licensed tool to edit code will not make the code itself copy-left licensed - see for example git itself.

Can‘t argue against tool duplication though. My scripts go back before pre-commit was popular and I missed Lucas-C‘s repo when updating them. So I guess we have more choice now :)

Thanks for looking at the PR though.

PR #677, opened 2022-07-04
build: add asort

cpendery:
## 📝 Description
Add support for `asort`, a python pip utility that keeps your `__all__` lists sorted.


asottile:
`require_serial` is wrong also and your pre-commit hook description doesn't help the user understand what your thing does

cpendery:
> you don't just get to jump the line -- add to the end like everyone else. and you absolutely don't need the emoji description. my initial assumption was this was spam!



> `require_serial` is wrong also and your pre-commit hook description doesn't help the user understand what your thing does

I'll update those and move it to the bottom. Thanks for the feedback

PR #678, opened 2022-07-04
build: add asort to pre-commit.com

cpendery:
## Description
Follow up to the previous PR addressing all the feedback given. Adds `asort` to the supported hooks list

Previous https://github.com/pre-commit/pre-commit.com/pull/677

my new repository:

- [ ] contains a license


asottile:
hi, I've edited in the new pr template to the original comment -- if you could fill that out as well

PR #680, opened 2022-07-05
build: add github-issue-prefix commit msg

KimSoungRyoul:
## Description
[add-msg-issue-prefix-hook](https://github.com/avilaton/add-msg-issue-prefix-hook) is used only in Jira

[add-github-issue-prefix](https://github.com/KimSoungRyoul/add-github-issue-prefix) support commit msg prefix for github issue and provides more customizable options.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/KimSoungRyoul/add-github-issue-prefix


asottile:
maybe I'm missing something, why do people keep adding "# Description" at the top of the description -- there's no reason to do that -- is there some tool that's doing this for you?

asottile:
- there's no license on this repo
- `always_run: true` is wrong
- did you try collaborating with the other repo first? I'd rather not have a bunch of slightly-different repos

KimSoungRyoul:
~~~
maybe I'm missing something, why do people keep adding "# Description" at the top of the description -- there's no reason to do that -- is there some tool that's doing this for you?
~~~
No there is no tool, I just thought it was a convention because  the previous person did it  lol :)

---

as is shown in pullrequest
https://github.com/avilaton/add-msg-issue-prefix-hook/pulls 
There is no feedback since 2021
so I made it

PR #682, opened 2022-07-08
Update all-repos.yaml

jeeftor:
Adding in the conventional commit hook

(https://github.com/compilerla/conventional-pre-commit)

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -217,5 +217,6 @@
 - https://github.com/Sceptre/sceptrelint
 - https://github.com/lyz-code/yamlfix
 - https://github.com/dannysepler/rm_unneeded_f_str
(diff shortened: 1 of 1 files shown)


asottile:
there's significantly more battle-tested tools in this space already on https://pre-commit.com/hooks.html (commitizen, etc.) -- I'd recommend collaborating with them rather than reimplementing.

we're also discouraging `language: system` / `language: script` repositories as it imposes installation requirements on the consumer which aren't the best user experience

additionally I would prefer if the repository isn't named "pre-commit" as it makes it confusing whether it's an official project or not.

thanks for the PR though!

jeeftor:
Fair enough - however commitzen has a really ugly fail message in comparison:

image

image

PR #684, opened 2022-07-14
Update all-repos.yaml

JanuszNowak:
Add https://github.com/Azure4DevOps/check-azure-bicep pre-commit for azure bicep

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/Azure4DevOps/check-azure-bicep


asottile:
- it's spelled "pre-commit" -- not "Pre-commit"
- `pass_filenames: true` is the default
- `language: script` hooks miss the point of the framework as they require users to install tools outside of the framework to use them and as such I'm not allowing new ones onto the hooks page

asottile:
> There is no support for powershell, you have hooks like for terraform that also require installation ?

...

>  not allowing new ones

JanuszNowak:
@asottile 
I fixed to `pre-commit`, also working on docker version. How can I use `additional_dependencies` to install all required stall like `pip install azure-cli`  ?

asottile:
you wouldn't -- you'd use `language: python` and use actual dependencies

PR #685, opened 2022-07-18
Add grintor-hooks to all-repos.yaml

grintor:
Added grintor-hooks repo to all-repos.yaml

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/grintor/grintor-hooks


grintor:
Hello,

I would like to add my hooks to your list of repos supporting pre-commit. One hook is a python port of the official AWS secrets detection bash script. The other is a tool for scanning for missing python dependencies in the requirements.txt file.

Thank you for your consideration,
Chris

all-repos.yaml
```diff
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/grintor/grintor-hooks
```
asottile, line 222: the first hook is missing proper license attribution (since it is a derivative work of the aws one) -- the second one doesn't work.  the pypi name doesn't have to match the import name

asottile: they often don't match so the design doesn't work

grintor: Oh! I misunderstood what you meant. Ok. I that's a great point. A clear oversight on my part. I need to determine the package corresponding to the import. I don't think that should be an insurmountable task. Allow me a few hours to research/test.

grintor: Alright, that was actually a pretty tough cookie to crack; but the latest published version solves the problem by actually parsing the pypi wheels from the requirements.txt file and determining the modules they expose. It can now do what it promises and reliably determine if you imported a module which is not installed by your requirements.txt.

asottile: hmmm now it has to run essentially a full `pip install` which is far too slow for a pre-commit hook (and won't work in network-isolated situations like pre-commit.ci)

grintor: The unfortunate situation is that the only way to determine if the requirements.txt is aligned with the projects is to examine the packages themselves as provided by the requirements.txt which does mean downloading them at least once (although if they have been installed on the system already, then they exist in the pip system cache which is shared with virtual environments, so there is no extra downloading required).

It can be pretty slow the first time while the building and then parsing of the wheels takes place. But I am hashing the requirements file and caching the data - so any subsequent runs are *extremely* fast so long as the requirements file doesn't change; it's a lot like pip-compile in that way.

As for the internet connection thing - As long as the packages have ever been installed on the system, then they will be in the pip system cache and not need to re-downloaded. There are lots of hooks that won't work without the packages having been installed - pylint for example would throw an error if you `import foo` and `foo` does not exist. So I think it's likely that most environments will not have this problem.

As for pre-commit.ci being network isolated, how does it deal with other hooks which require internet connectivity like python-safety-dependencies-check? And how does it deal with situations like the pylint example above?

asottile: they are not allowed in pre-commit-ci and generally discouraged. you shouldn't need an internet connection to run linters and formatters. things which require that are best left to your test suite

grintor: I see. I can respect that. I do wish there were a way to make detect-missing-requirements work without internet, but I think it's fundamentally impossible to do this since the source of truth is located in the pypi repo and constantly changing.

I at least think people will find detect-missing-requirements useful - even if it does not meet the purest guidelines of pre-commit - I find using using pylint with pre-commit very useful even if it's discouraged.

At least the other hook in my repo meets with your approval.

PR #688, opened 2022-07-22
Add cpp-linter-hooks to the all-repos list

shenxianpeng:
Thanks for creating such a great framework.

I know there are already hooks that support `clang-format` and `clang-tidy`, but I still want to recommend this hook because:

1. It supports users to use any specific versions of `clang-format` and `clang-tidy` (from v3.9 to v14)
2. `clang-format` and `clang-tidy` binares can be automatically installed on Windows, Linux, and macOS if they do not exist.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/dannysepler/rm_unneeded_f_str
 - https://github.com/cmhughes/latexindent.pl
 - https://github.com/sirwart/ripsecrets
+- https://github.com/cpp-linter/cpp-linter-hooks


asottile:
this has a few problems that the existing things don't
- the tools are downloaded at runtime rather than install time (can't be cached properly, can't be used in pre-commit.ci)
- files and types_or are both specified (almost always a bug)

shenxianpeng:
Thanks for the review.

* If the tools don't exist on the system, they will only be downloaded once at runtime. if the tools are installed at install time, they will not be installed at runtime. (by default, the tools are installed to the path `~/.local/bin/`, if pre-commit.ci uses the same user id, it can be shared.)
* I have removed `types_or` only left `files`, and passed the test.

asottile:
pre-commit.ci has no internet access at runtime -- if you need to install things they need to be installed at install time

shenxianpeng:
OK got it, used in pre-commit.ci needs to be pre-installed. I think in most cases, the user does not need to care whether the tool is installed or not, this hook will automatically solve it

asottile:
yeah the other versions of this do better by actually performing the install during the install -- they provide a better user experience and can be used in more places

shenxianpeng:
Using `clang-format` and `clang-tidy` on Windows and other Linux distributions except for Ubuntu (maybe also on macOS), it's inconvenience to get and install `clang-format` and `clang-tidy` executables, because [llvm-project](https://github.com/llvm/llvm-project/releases) doesn't provide them directly, users need manually prepare when they using the current supported hooks, so I create this one to solve the problem.

It's OK to merge it or not. thanks for the comments.

asottile:
check out how https://github.com/pre-commit/mirrors-clang-format works -- works on windows and installs at install-time

shenxianpeng:
I saw this repo, it only supports clang-format, it would be nice if it could support clang-tidy.
The existing hook which supports clang-tidy is https://github.com/pocc/pre-commit-hooks, but it needs to manually install tools before runtime.

PR #691, opened 2022-07-27
Add compilerla/conventional-pre-commit to all-repos.yaml

thekaveman:
This closes compilerla/conventional-pre-commit#24.

~@asottile I saw [your comment in #684](https://github.com/pre-commit/pre-commit.com/pull/684#issuecomment-1184377808)~

~> `language: script` hooks miss the point of the framework as they require users to install tools outside of the framework to use them and as such I'm not allowing new ones onto the hooks page~

~And while `conventional-pre-commit` is a `language: script` hook, it's literally just a [small shell script](https://github.com/compilerla/conventional-pre-commit/blob/main/conventional-pre-commit.sh) that does some regex against the commit message, no external tools/dependencies installed or needed. Maybe an exception can be made in this case?~

Python hook that checks for [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format, intended to run in the `commit-msg` stage.

Thanks for taking a look!

___

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -230,3 +230,4 @@
 - https://github.com/jshwi/docsig
 - https://github.com/finsberg/clang-format-docs
 - https://github.com/rubocop/rubocop
+- https://github.com/compilerla/conventional-pre-commit


asottile:
> This closes [compilerla/conventional-pre-commit#24](https://github.com/compilerla/conventional-pre-commit/issues/24).
> 
> @asottile I saw [your comment in #684](https://github.com/pre-commit/pre-commit.com/pull/684#issuecomment-1184377808)
> 
> > `language: script` hooks miss the point of the framework as they require users to install tools outside of the framework to use them and as such I'm not allowing new ones onto the hooks page
> 
> And while `conventional-pre-commit` is a `language: script` hook, it's literally just a [small shell script](https://github.com/compilerla/conventional-pre-commit/blob/main/conventional-pre-commit.sh) that does some regex against the commit message, no external tools/dependencies installed or needed. Maybe an exception can be made in this case?
> 
> Thanks for taking a look!

it appears to require a modern `bash` as well as `grep` -- it would be better if implemented in python -- or really the whole script can be replaced with a [`pygrep`](https://pre-commit.com/#pygrep) regex

thekaveman:
Hello @asottile - since last week, `conventional-pre-commit` has been reimplemented in Python and a new version `v2.0.0` published. Thank you for the suggestion!

Does it look OK to include in the list now?

asottile:
is there a reason to use this over something more mature like `gitlint` ?

asottile:
hi -- I've edited the new pr template into the original comment -- please fill that out as well

PR #692, opened 2022-07-29
Add pre-commit-checkstyle to all-repos.yaml

junghoon-vans:
`Checkstyle` is a static code analysis tool for checking Java coding conventions.
`pre-commit-checkstyle` is to easily run checkstyle command.


asottile:
- please do not use the name "pre-commit" when not associated with the project
- we are no longer accepting `language: script` / `language: system` hooks on the page because they do not work without the user needing to install things outside of the framework (missing the main point of the framework)

PR #698, opened 2022-08-12
Update pre-commit-hooks version in Quick start

keyem4251:
I try pre-commit Quick start and notice pre-commit-hooks version is not latest.
So update pre-commit-hooks version in Quick start.


asottile:
dupe #632

PR #701, opened 2022-08-20
Add pre-commit’s `validate_manifest` hook

Jayman2000:


asottile:
lol you think I would forget this? it's intentionally not listed because it's just an example

Jayman2000:
> it's intentionally not listed because it's just an example

Wait, so is it not useful? I’m asking because I have it enabled for [this repo](https://github.com/Jayman2000/pre-commit-mkdocs).

asottile:
not really, pre-commit always validates the config when used

PR #702, opened 2022-08-20
Add actionlint to pre-commit list

yoichi:
[actionlint](https://github.com/rhysd/actionlint) is a static checker for GitHub Actions workflow files.


asottile:
`language: system` hooks aren't being accepted any more -- it should use `language: go` since it's written in a supported pre-commit language already -- btw there's also https://github.com/python-jsonschema/check-jsonschema/blob/main/.pre-commit-hooks.yaml

PR #703, opened 2022-08-22
Add hooks for `shellcheck` & `vale`

Jarmos-san:
Add a couple more hooks. More information are as follows:

- `shellcheck-precommit` is an improved unofficial version of [koalaman/shellcheck-precommit](https://github.com/koalaman/shellcheck-precommit). The improved version doesn't require a build phase & instead uses the installed `shellcheck` binary.
- `vale-precommit` is a linter for checking grammatical/spelling errors in prose like content written in Markdown & Plan Text files.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -221,3 +221,5 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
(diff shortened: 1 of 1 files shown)


asottile:
- there's already a shellcheck that's maintained by me and is better than the one you're suggesting
- `language: system` is also not allowed for new hooks as it misses the point of the framework
- please don't use "pre-commit" in the naming of your tools as you are not associated with the project

PR #704, opened 2022-08-23
Add phylum-ci

maxrake:
The `phylum-ci` hook uses [phylum](https://www.phylum.io) to provide analysis of project dependencies from a lockfile during a commit containing that lockfile. The hook will fail and provide a report if any of the newly added/modified dependencies from the commit fail to meet the project risk thresholds for any of the five Phylum risk domains:

* Malicious Code
* Software Vulnerabilities
* Authorship Risk & Reputation
* License Misuse
* Engineering Risk

See [Phylum Risk Domains documentation](https://docs.phylum.io/docs/phylum-package-score#risk-domains) for more detail.

The hook will be skipped if no dependencies were added or modified for a given commit. If one or more dependencies are still processing (no results available), then the hook will only fail if dependencies that have completed analysis results do not meet the specified project risk thresholds.

Relevant links for `phylum-ci`:
* [GitHub repo](https://github.com/phylum-dev/phylum-ci)
* [Hook config](https://github.com/phylum-dev/phylum-ci/blob/main/.pre-commit-hooks.yaml)
* [Documentation for using the pre-commit hook](https://github.com/phylum-dev/phylum-ci/blob/main/docs/sync/git_precommit.md)
* [PyPI package](https://pypi.org/project/phylum/)
* [The PR where the hook was added](https://github.com/phylum-dev/phylum-ci/pull/91)
  * There are good screenshots here to show how the hook looks/works
  * Similar results can be obtained by using known bad packages
    * `pyyaml==5.3.1` is one such package

Links for Phylum:
* [Phylum main page](https://www.phylum.io/)
* [Phylum's documentation](https://docs.phylum.io/docs/welcome)
* [Register for the free Community Edition](https://app.phylum.io/register)

**NOTE**: Adding this hook here was purposefully delayed until the [Phylum Community Edition was released](https://www.prnewswire.com/news-releases/phylum-releases-a-free-community-edition-to-make-software-supply-chain-security-more-accessible-301599698.html). This is a free version that allows for up to five (5) projects. Using this hook will allow developers to shift their security further left, ensuring bad open source software dependencies are not committed to their repository.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/phylum-dev/phylum-ci


all-repos.yaml
```diff
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/phylum-dev/phylum-ci
```
asottile, line 224: this doesn't appear to be configured correctly:

- it does not filter filenames
- it sets `require_serial` though afaict doesn't involve any sort of multiprocessing

maxrake: > * it does not filter filenames

This is on purpose. Phylum supports analyzing dependency lockfiles from a range of languages. Some lockfile types (e.g., Python/pip `requirements.txt`) are ambiguous in that they can be named differently (e.g., `requirements-dev.txt`, `dev-requirements.txt`, etc.) and may or may not contain strict dependencies (e.g., `pyyaml` vs. `pyyaml==5.3.1`). In these cases, it is best to specify an explicit lockfile path.

Instead of trying to capture all possible lockfile names in the `.pre-commit-hooks.yaml` file, the [documentation to configure `.pre-commit-config.yaml`](https://docs.phylum.io/docs/git_precommit#configure-pre-commit-configyaml) specifies the [file control/filter method](https://docs.phylum.io/docs/git_precommit#file-control) to limit the hook being run to only the times when the explicitly identified lockfile has changed, saving execution time.

Further, the list of currently supported lockfile types, which is expanding as new languages are supported, can be viewed [in documentation](https://docs.phylum.io/docs/analyzing-dependencies) or [in source](https://github.com/phylum-dev/phylum-ci/blob/main/src/phylum/constants.py#L31).

> * it sets `require_serial` though afaict doesn't involve any sort of multiprocessing

Upon further review, `require_serial` is not required to be set for the `phylum` hook. However, changing that meant also changing how the script/hook exits. These changes are currently in the [`not_serial` branch](https://github.com/phylum-dev/phylum-ci/tree/not_serial) of the repository and will be merged to `main` after a review...likely this week.

PR #705, opened 2022-08-23
Add conhead consistent headers pre-commit hook

slobberchops:
conhead (short for consistent headers) is a tool that is useful for adding headers to files (typically license headers) and keeping them up to date (maintaining copyright years over time). It is available as a pre-commit hook.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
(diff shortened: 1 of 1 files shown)


all-repos.yaml
```diff
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/slobberchops/conhead
```
asottile, line 224: it probably makes sense to collaborate with existing and more battle-tested implementations of this such as https://github.com/Lucas-C/pre-commit-hooks#insert-license

PR #706, opened 2022-08-24
Add checkstyle-cli

junghoon-vans:
`checkstyle-cli` is a command-line tool for checkstyle(static analysis tool for java)

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/junghoon-vans/checkstyle-cli


all-repos.yaml
```diff
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/junghoon-vans/checkstyle-cli
```
asottile, line 224: the download would be better if done at build time rather than runtime (ala https://github.com/shellcheck-py ) that way it would be compatible with pre-commit.ci 

additionally there's already coursier support -- can checkstyle be installed directly with that? might be a better approach

junghoon-vans: I was looking for a way to apply it to pre-commit.ci, but thank you for letting me know.

I think there are two problems with using the courseier.
First of all, if you are using it locally, you cannot run it unless the coursier is installed. Most Java developers do not have the coursier installed locally.
Second, the courseier is not yet supported on pre-commit.ci.

PR #707, opened 2022-08-26
Add buildifier pre-commit hook

Warchant:
Buildifier is a linter/formatter for Bazel files (.bzl, BUILD, WORKSPACE).

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/Warchant/pre-commit-buildifier


all-repos.yaml
```diff
@@ -221,3 +221,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/Warchant/pre-commit-buildifier
```
asottile, line 224: we're no longer accepting new `language: script` hooks -- sorry!

Warchant: But it's python only hook. 
https://github.com/Warchant/pre-commit-buildifier/blob/main/.pre-commit-hooks.yaml#L5

asottile: ah I see -- it still downloads at runtime which isn't ideal

there's also several other buildifier tools listed on the page already -- I'd recommend collaborating

also please don't use "pre-commit" in your name since you're not affiliated with the project

PR #710, opened 2022-09-08
clarify autoupdate selection logic

eviltwin:
When reporting an issue on pre-commit earlier, I had misunderstood what the selection logic was for the "latest" tag to be updated to. When I was directed to the manual, it wasn't immediately apparent to me what I had misunderstood. I suspect that this is because the manual is clear and concise to the repository maintainers who are familiar with the code but doesn't give enough context to those of us who are just mere users of the tool.

I'd like to propose the following additions to the `autoupdate` section of the docs, to help clarify the logic of which tag is chosen. Hopefully this change will result in fewer erroneous bug reports and also provide more clarity to people who are directed to the docs in response to a github issue.

I considered adding a note about the fact that git tag dates are not available within git repositories (they're a feature of github and its api) as I had found this confusion from another user in a different issue... but I chose to leave it out as I felt that the docs can't possibly answer _every_ permutation of the question and it was better to keep things brief for the general case.

### DIFF
--- a/sections/cli.md
+++ b/sections/cli.md
@@ -15,9 +15,12 @@ _new in 2.8.0_: `pre-commit` now exits with more specific codes:
 
 Auto-update pre-commit config to the latest repos' versions.
 
+By default, pre-commit will choose the latest tag on the repository's default
+branch, preferentially picking a tag that contains a `.` in the case of ties.
+
 Options:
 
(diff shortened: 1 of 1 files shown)


asottile:
I don't think this improves anything -- and it adds another place where people will send annoying PRs to update versions

the help text is pretty clear that it's the default branch already

PR #712, opened 2022-09-13
Update install.md

denadai2:
I propose to update the hooks version, but especially black: the specified version has a bug and create issues with new people trying to follow the steps installing pre-commit.

ref: https://github.com/psf/black/issues/2964


asottile:
it's not meant to be kept up to date

PR #714, opened 2022-09-15
Add `fabasoad/pre-commit-snyk` hook

fabasoad:
Add `fabasoad/pre-commit-snyk` hook to `all-repos.yaml`.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -220,3 +220,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/fabasoad/pre-commit-snyk


all-repos.yaml
```diff
@@ -220,3 +220,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/fabasoad/pre-commit-snyk
```
asottile, line 223: a few things:

- this would be better done upstream I think
- please don't use "pre-commit" when naming your project, you aren't associated with pre-commit
- we're no longer accepting new `language: system` / `language: script` hooks which require the user to manually install things outside of the framework (it misses the point of the framework and creates a negative user experience)

PR #716, opened 2022-09-15
Add qarik-group/gitlab-codeowners-linter

gasparev:
`qarik-group/gitlab-codeowners-linter` is a linter for [Gitlab's CODEOWNERS file](https://docs.gitlab.com/ee/user/project/code_owners.html). Note that the syntax for Gitlab is different than Github's one so linters like `github.com/milin/gitown` are not compatible.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -219,3 +219,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/Qarik-Group/gitlab-codeowners-linter


all-repos.yaml
```diff
@@ -219,3 +219,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/Qarik-Group/gitlab-codeowners-linter
```
asottile, line 222: https://github.com/Qarik-Group/gitlab-codeowners-linter/blob/901029be1587bf9a005efe1688fb70447033b102/.pre-commit-hooks.yaml#L7-L8

- this regex should have a `$`
- `pass_filenames: false` doesn't gel nicely with the framework -- your tool should handle filenames passed as positional arguments

gasparev: I fixed the 2 points you mentioned

asottile: the second was not fixed -- your tool should handle a variable number of positional arguments and lint those files

gasparev: That's not a valid use case, there can be [only 1 COEOWNERS file per repo](https://docs.gitlab.com/ee/user/project/code_owners.html#set-up-code-owners)

asottile:
I'm not going to argue with you

all-repos.yaml
gasparev: I modified the tool, it can now handle a variable number of positional arguments and lint those files

PR #717, opened 2022-09-16
docs: add claylo/commit-msg-hooks

claylo:
I hope these hooks will fill some gaps in robust & flexible commit-msg hook usage!

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -220,3 +220,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/claylo/commit-msg-hooks


all-repos.yaml
```diff
@@ -220,3 +220,4 @@
 - https://github.com/sirwart/ripsecrets
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
+- https://github.com/claylo/commit-msg-hooks
```
asottile, line 223: what does this bring that a more battle-tested tool like `gitlint` doesn't?  also the `regex` thing is completely obsoleted by a builtin pre-commit functionality (`pygrep`)

asottile (commented):
we don't do conventional commits here

all-repos.yaml
claylo: `pygrep` - the vast majority of those hooks specify they're for acting on python files, not commit messages.
https://github.com/pre-commit/pygrep-hooks/blob/main/.pre-commit-hooks.yaml I don't see how these hooks are remotely similar to what my two hooks are doing. Are you referring to some other pygrep?

`gitlint` - I overlooked that in the list, and wasn't familiar with it otherwise. 

The main difference with gitlint and the hooks I've submitted is that mine don't require a completely separate tool. In my experience, people aren't looking for a deep toolchain to manipulate their commit messages -- they want to apply a couple of regular expressions and be done. That said, I'll explore gitlint further.

`msgcheck` is lightweight (under 200 lines of bash) and that's it.

`trailers` differs from both gitlint and msgcheck in that is operates solely on the output of `git interpret-trailers --only-trailers`, so none of the regular expressions defined will inadvertently match any portion of a commit message that git itself does not consider a trailer.


Suggestion: it's more likely I'd have seen gitlint if hooks were organized somewhere on the pre-commit site by hook-type.

Also: is there a new policy about duplicate functionality that I overlooked? I assumed from the [list of supported hooks](https://pre-commit.com/hooks.html) that hooks implementing a particular behavior in a different way are welcomed.

asottile: for pygrep I'm referring to `language: pygrep` -- a "language" for implementing simple regex-based hooks

also just a heads up your README's sponsorship looks spammy / suspect

claylo: > there isn't -- but I'd prefer to stick to things that are already there
> 
> also I noticed your tool is written with `language: script` / `language: system` which we're no longer accepting on the list due to requiring specific setup on the user's side (the point of pre-commit is that it manages the tools avoiding the need for the user to set up things)

So ... bash? You're not taking entries written in bash?

You really need to update your site to reflect that. There's nothing there stating that writing a hook in bash for inclusion on the site is a complete waste of time. That would have been good to know in advance.

PR #724, opened 2022-09-30
Add MacPorts installation instructions

ryandesign:
You can [install pre-commit using MacPorts](https://ports.macports.org/port/pre-commit/)! This PR adds this information to the install documentation.


asottile:
thanks! though this list isn't meant to be exhaustive

PR #729, opened 2022-10-17
Add python-unittest

sharkwouter:
This is a very simple hook which runs `python3 -m unittest discover -qb`. The q and b are to silence any output a successful test might have, but errors are still shown.

The code for the hook can be found here: https://github.com/sharkwouter/python-unittest

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -220,3 +220,4 @@
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
 - https://github.com/daveshanley/vacuum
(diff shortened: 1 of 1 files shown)


asottile:
tests should not be run in pre-commit: the repository under test is not installed and they are far too slow for a pre-commit hook and will likely lead to your contributors turning off the entire framework

sharkwouter:
I'm surprised by that being your experience. Unit test should be almost instant.

Do you need help with cleaning up other unit test runners from the website like the ones for Java, Go and Kotlin which are listed there?

PR #731, opened 2022-10-18
Update all-repos.yaml

JanuszNowak:
Add https://github.com/Azure4DevOps/check-azure-bicep with support dependencies using python.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -220,3 +220,4 @@
 - https://github.com/bagerard/graphviz-dot-hooks
 - https://github.com/omnilib/ufmt
 - https://github.com/daveshanley/vacuum
+- https://github.com/Azure4DevOps/check-azure-bicep


asottile:
`additional_dependencies` is usually wrong in `.pre-commit-hooks.yaml` -- we're also no longer accepting new `language: script` hooks as they require manual setup for users outside of the framework

also it seems like this runs a build? that's probably not a good candidate for pre-commit hooks which should be very fast checks (or they'll discourage users leading to them turning off the entire framework)

JanuszNowak:
Hi @asottile 
there are 2 hooks one using script that allow to have local dependencies already installed, second one installation dependencies, so you can choose what work best for you. Execution of it is like 3-4 sec
https://raw.githubusercontent.com/Azure4DevOps/check-azure-bicep.example/master/example.gif

- id: check-azure-bicep
  language: script

- id: check-azure-bicep-python
  language: python
  additional_dependencies: ['azure-cli']

asottile:
yeah that's far too slow for a pre-commit hook

JanuszNowak:
@asottile bunch of pre-commit already there are working longer, but are we are forgetting that we scan only modified files ? Like editorconfig pre-commit is working sometimes for 10 sec depending on form volume, and if it is executing too long for somebody he/she is not forced to use it ? More or less, this is for awareness. Somebody can find it if needed and he/she accept time. Can I deliver some benchmark ? If somebody is working with tool, he/she already accepting time. In this is pre request then propalby all pre-commit hooks executing unit tests suppose to be removed as they can execute for multiple minutes depending on from volume.

PR #736, opened 2022-10-26
Add CircleCI hook

zahorniak:
This hook validates CircleCI config using circleci cli


asottile:
new `language: script` / `language: system` hooks are no longer being allowed because they miss the point of the framework and require manual installation instructions outside of pre-commit

PR #742, opened 2022-11-13
Add poetry-audit to check poetry.lock for vulnerabilities

koyeung:
Adding poetry-audit hook to page https://pre-commit.com/hooks.html.


asottile:
this is far too slow to be a pre-commit hook. plus poetry already provides hooks so you don't need a separate repo

PR #745, opened 2022-11-16
add pre-commit-trailer and pre-commit-ssh-git-signing-key

adam-moss:
adds pre-commit hooks to
- add a Pre-Commit-Config: trailer to commit messages on `commit-msg`
- add ssh based git signing key to ssh agent on `commit-msg`
- remove ssh based git signing key from ssh-agent on `post-commit`

Implemented as POSIX shell scripts for portability and simplicity.

In https://github.com/pre-commit/pre-commit.com/pull/684#issuecomment-1184377808 you state

> - `language: script` hooks miss the point of the framework as they require users to install tools outside of the framework to use them and as such I'm not allowing new ones onto the hooks page

I have raised this merge request for visibility and considerations of the `add-pre-commit-config-trailer` hook, which was mentioned in https://github.com/pre-commit/pre-commit/issues/582#issuecomment-1316032165 as a solution to an issue outstanding since 2017.

Anyone wanting to utilise these hooks would by definition already have the relevant tooling (`git` and `ssh`) installed so abstracting into python would simply increase the lines of code and associated dependency tree for no material gain.

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -222,3 +222,5 @@
 - https://github.com/omnilib/ufmt
 - https://github.com/daveshanley/vacuum
 - https://github.com/nuztalgia/botstrap
+- https://gitlab.com/adam-moss/pre-commit-trailer
+- https://gitlab.com/adam-moss/pre-commit-ssh-git-signing-key


all-repos.yaml
```diff
@@ -222,3 +222,4 @@
 - https://github.com/omnilib/ufmt
 - https://github.com/daveshanley/vacuum
 - https://github.com/nuztalgia/botstrap
+- https://gitlab.com/adam-moss/pre-commit-hooks
```
asottile, line 225: shell limits the portability, especially for windows -- it would be preferable to implement these in one of the supported languages

also could you rename your repo to something more specific?  repos named `pre-commit-hooks` are easy to confuse with the framework and it'd be easier to find with a name specific to what you've written

adam-moss: In terms of language I'm not look to re-implement these hooks at this time.  I know there may be issues under WSL but I don't have access to Windows based device to test efficiently or regularly, so any support would only every be on a reasonable endeavours basis.

adam-moss: Renamed (and split) repo as requested.

Thanks

adam-moss:
Added a complimentary hook to `pre-commit-trailer` so alongside the checksum of the `.pre-commit-config.yaml` file you can also have a `Pre-Commit-User-Skipped:` trailer for any hooks by-passed by use of the `$SKIP` environment variable.

PR #749, opened 2022-11-22
adding pre-commit-sync hook

miki725:
that hook allows to sync pre-commit versions from lock files

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -222,3 +222,4 @@
 - https://github.com/omnilib/ufmt
 - https://github.com/daveshanley/vacuum
 - https://github.com/nuztalgia/botstrap
+- https://github.com/crashappsec/pre-commit-sync


miki725:
> it requires network

what requires network? the hook gets the versions from the lock files directly from the file system. no network is being used at all. can you elaborate what you mean?

> it's far too slow to use in the general case, your users will get frustrated and turn off the whole framework

what do you mean too slow in general case? by default it only runs when any of the lock file or precommit config files change:

https://github.com/crashappsec/pre-commit-sync/blob/316c8cedfabf6f8659a63d205ded338538e32876/.pre-commit-hooks.yaml#L9-L15

only overhead is parsing supported lock files which is relatively quick:

```
➜ time python -m pre_commit_sync poetry.lock > /dev/null

________________________________________________________
Executed in  150.51 millis    fish           external
   usr time  120.23 millis  246.00 micros  119.98 millis
   sys time   29.93 millis   26.00 micros   29.90 millis
```

> it's not licensed

that was an oversight. it now has Apache 2.0 license.

https://github.com/crashappsec/pre-commit-sync/blob/main/LICENSE

> please don't use the name pre-commit since your hook is not official in any way

there are numerious other non-official pre-commit hooks in `all-hooks.yml` with `pre-commit-` prefix

https://github.com/pre-commit/pre-commit.com/blob/main/all-repos.yaml

if its a concern README can clearly state its not an official hook

miki725:
@asottile any comments on the above response? I think the hook was dismissed due to a misunderstanding. happy to answer any questions you might have

PR #750, opened 2022-11-27
Add 'pre-commit-env-sample-sync' plugin

acaloiaro:


all-repos.yaml
```diff
@@ -224,3 +224,4 @@
 - https://github.com/nuztalgia/botstrap
 - https://gitlab.com/adam-moss/pre-commit-trailer
 - https://gitlab.com/adam-moss/pre-commit-ssh-git-signing-key
+- https://github.com/acaloiaro/pre-commit-env-sample-sync
```
asottile, line 227: please don't name your thing with "pre-commit" in the name -- it's not officially associated

we're no longer accepting `language: script` / `language: system` hooks as they're not portable

PR #754, opened 2022-12-07
Add terraform-docs

william-richard:
I just found these hooks and want to add them to the list 😸

my new repository:

- [ ] contains a license


davidjb:
I'm not the OP, but was looking at adding this also:



my new repository:

- [x] contains a license
- [ ] is not `language: system` or `language: docker` (3 hooks are in the repo, 1 `golang`, 1 `system` and 1 `docker`)

asottile:

PR #756, opened 2022-12-11
Adds checkmake repo

colindean:
checkmake got a hook configuration in https://github.com/mrtazz/checkmake/commit/4e37f3a6ea00208b3f5bd04eaa0f075068237f18

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -225,3 +225,4 @@
 - https://gitlab.com/adam-moss/pre-commit-trailer
 - https://gitlab.com/adam-moss/pre-commit-ssh-git-signing-key
 - https://github.com/charliermarsh/ruff-pre-commit
+- https://github.com/mrtazz/checkmake


asottile:
`pass_filenames: false` is incorrect I believe -- the tool should operate on files passed to it

colindean:
Where are you seeing that?

Looks like it's set to `true` [here](https://github.com/mrtazz/checkmake/blob/4e37f3a6ea00208b3f5bd04eaa0f075068237f18/.pre-commit-hooks.yaml#L5).

asottile:
ah I think I just misread -- `pass_filenames: true` is the default so you shouldn't need to set it

is there a reason that `types: [makefile]` wasn't used?

colindean:
Looks like `*.make`, declared [here](https://github.com/mrtazz/checkmake/commit/4e37f3a6ea00208b3f5bd04eaa0f075068237f18#diff-cb921d5df435fe404b4daf451cf783709ccebe9c346038af6036644e7386c13dR10), is missing from the extension list [here](https://github.com/pre-commit/identify/blob/e834194e383eb683ad73a1d0324aaf66942edc8c/identify/extensions.py#L137) but the filenames are listed [here](https://github.com/pre-commit/identify/blob/e834194e383eb683ad73a1d0324aaf66942edc8c/identify/extensions.py#L321-L328) adequately.

I've proposed adding `.make` in https://github.com/pre-commit/identify/pull/353 and switching to the `types` property in https://github.com/mrtazz/checkmake/pull/77. It might be a hot minute before 353 makes it into a pre-commit release but I think that's OK because I've rarely seen `.make` and the change in checkmake is so recent that I doubt that anyone who's adopted the pre-commit hook already relies on it.

asottile:
identify has been released so you should be able to use `types` now

PR #761, opened 2022-12-20
TypeScript compiler checker

ggozad:
…er check

my new repository:

- [X] contains a license
- [X] is not `language: system` or `language: docker`

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -226,3 +226,4 @@
 - https://gitlab.com/adam-moss/pre-commit-ssh-git-signing-key
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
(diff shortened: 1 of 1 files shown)


asottile:
you've checked the 4th checkbox incorrectly

asottile:
and calling `npx tsc` escapes the framework so there's no benefit to this over `language: system`

ggozad:
Hey Anthony, very sorry about this, wasn't paying enough attention. 
I renamed the repo and  removed npx so will  make another pull request.

asottile:
I don't think tsc is a good fit for pre-commit, it's going to be too slow *and* it requires your dependencies be available

ggozad:
Fair enough :)
I thought I would make this available as I could not find an alternative hook for type checking. This is a check that I would guess is very common in js/ts projects. Out of curiosity is there an alternative faster way?

PR #762, opened 2022-12-25
add docsig

jshwi:
check signature params for proper documentation

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -226,3 +226,4 @@
 - https://gitlab.com/adam-moss/pre-commit-ssh-git-signing-key
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
+- https://github.com/jshwi/docsig


all-repos.yaml
```diff
@@ -226,3 +226,4 @@
 - https://gitlab.com/adam-moss/pre-commit-ssh-git-signing-key
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
+- https://github.com/jshwi/docsig
```
asottile, line 229: why does this use `require_serial: true` ?

jshwi: Hi @asottile,
Thank you for checking this
I'm new to `pre-commit`, I sort of inherited from a hook I was learning from
Come to think of it the hooks I've seen since don't have that, removed

asottile (approved):

jshwi:
Thanks @asottile! 😄

PR #764, opened 2022-12-29
Add more hooks for Elixir

ahamez:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -227,3 +227,4 @@
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
 - https://github.com/jshwi/docsig
+- https://github.com/ahamez/pre-commit-elixir-hooks


ahamez:
Hello,

What can I do to have this PR accepted? Are all the items in the checklist mandatory? If so, how can I add hooks for a language that is not in the list of [supported languages][(url](https://pre-commit.com/#supported-languages))?

Thank you :-)

asottile:
why would there be a checklist if they aren't mandatory?

ahamez:
TBH, as there were hooks that contained `pre-commit` [in](https://github.com/pre-commit/pre-commit.com/blob/0506f4320e7811307bfa9a43458511e557474819/all-repos.yaml#L228) [the](https://github.com/pre-commit/pre-commit.com/blob/0506f4320e7811307bfa9a43458511e557474819/all-repos.yaml#L207) [name](https://github.com/pre-commit/pre-commit.com/blob/0506f4320e7811307bfa9a43458511e557474819/all-repos.yaml#L203) or given the fact that nothing in the documentation about [`system`](https://pre-commit.com/#system) states it should not be used, it was not that obvious :-).

While I can easily change the name, I don't see how I could add a language which is not the list of [supported languages](https://pre-commit.com/#supported-languages) if one is not allowed to use `system`?

ahamez:
@asottile Hello again 🙂 So, what do you recommend if I can't use `system` nor `docker` ?  [`script`](https://pre-commit.com/#script)?
Also, could I suggest that you explicitly write in the [documentation](https://pre-commit.com/#supported-languages) that both `system` and `docker` are not allowed?

PR #768, opened 2023-01-02
add isops, a sops secrets checker

lorenzophys:
My first contribution to the pre-commit community:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -227,3 +227,4 @@
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
 - https://github.com/jshwi/docsig
+- https://github.com/lorenzophys/isops


asottile:
- `pre-commit` is misspelled -- it is not capitalized
- `pass_filenames: false` misses the point of the framework, tools should take positional arguments of files to check

lorenzophys:
Hey @asottile I see your point, although when dealing with sops the list of files to check is determined by the sops config file: it's not possible to know what files need to be checked when running the command without first parsing the config file.

Of course you can just list them manually when you configure the hook, but a tool like this is needed exactly because you might forget one of the files that you need to check before `git commit`. :D 

Does this make sense?

asottile:
it's not acceptable on the hooks page if it doesn't gel nicely with the framework so I can't accept it in the current state

lorenzophys:
Fair enough! I'm closing

PR #772, opened 2023-01-08
add yamlfmt to all-repos

braydonk:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -165,6 +165,7 @@
 - https://github.com/yoheimuta/protolint
 - https://github.com/hadolint/hadolint
 - https://github.com/google/go-jsonnet
+- https://github.com/google/yamlfmt
 - https://github.com/guid-empty/flutter-dependency-validation-pre-commit
 - https://github.com/cpplint/cpplint
 - https://github.com/MarcoGorelli/absolufy-imports


all-repos.yaml
```diff
@@ -227,3 +227,4 @@
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
 - https://github.com/jshwi/docsig
+- https://github.com/google/yamlfmt
```
asottile, line 230: `yamlfmt .` is going to double-run every file (pre-commit works by passing the files as positional arguments to hooks).  there are two choices: (1) remove the `.` (recommended) or (2) set `pass_filenames: false` (I'm not accepting new things to the list which sets this)

braydonk: Makes sense, I'll remove that and re-open this PR on the next release.

braydonk:
I'm reopening this PR for consideration. I made the recommended adjustment by removing `.` from `entry`.

all-repos.yaml
djgoku: Looks like `v0.8.0` was tagged and removes the `.` [1].

[1] https://github.com/google/yamlfmt/blob/v0.8.0/.pre-commit-hooks.yaml

asottile: nice!  can you move this up next to the other `google/` hook?

braydonk: Ah good idea, done in [afeac4c](https://github.com/pre-commit/pre-commit.com/pull/772/commits/afeac4c645a9276353d037f65e87807f01670c84)!

asottile (approved):

PR #775, opened 2023-01-18
Add zeek-format to to list of all hooks

AmazingPP:
[zeek-format](https://github.com/zeek/zeekscript#zeek-format) is a [Zeek script](https://docs.zeek.org/en/master/scripting/basics.html) formatter.


asottile:
don't skip the issue template

PR #777, opened 2023-01-22
add constcheck

jshwi:
check python files for repeat use of strings

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -227,5 +227,6 @@
 - https://github.com/charliermarsh/ruff-pre-commit
 - https://github.com/mrtazz/checkmake
 - https://github.com/jshwi/docsig
+- https://github.com/jshwi/constcheck
 - https://github.com/finsberg/clang-format-docs
 - https://github.com/rubocop/rubocop


asottile:
you seem to misunderstand pre-commit -- it will never check files which aren't checked in

asottile:
also this tool is probably better done as a flake8 or pylint plugin rather than a standalone tool

jshwi:
> you seem to misunderstand pre-commit -- it will never check files which aren't checked in

Hey @asottile,
I'm a little lost on this one, I'm aware of that, not sure why I appear to misunderstand this

jshwi:
> also this tool is probably better done as a flake8 or pylint plugin rather than a standalone tool

Yes, I have considered that, should I come back once this is integrated?

asottile:
> > also this tool is probably better done as a flake8 or pylint plugin rather than a standalone tool
> 
> Yes, I have considered that, should I come back once this is integrated?
> 
> Although, it only just shows a tally and returns a non zero exit status on failure, so not sure where this would fit in with the error categories you see with flake8 or pylint

once it's a plugin for those it's not listed here -- it's just something one would use with those

PR #778, opened 2023-01-23
add dbt-checkpoint to list of repos

MartinGuindon:
my new repository:

- [X] contains a license
- [X] is not `language: system` or `language: docker`

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -229,3 +229,4 @@
 - https://github.com/jshwi/docsig
 - https://github.com/finsberg/clang-format-docs
 - https://github.com/rubocop/rubocop
+- https://github.com/dbt-checkpoint/dbt-checkpoint


asottile:
it seems the docs still have the old name?

MartinGuindon:
We're in the process of cleaning that up! This project was more or less abandoned, and we ended up with two organizations who forked the project, willing to take it over. When we realized that, we teamed up to co-manage the project. We are in the process of merging our (diverging) changes together into a new version, and cleaning up the docs and code base to reflect the new name.

There is a redirect from the old repository name so its not breaking. When we're done with the cleanup, it should reduce the confusion.

asottile:
let me know when the dust settles and then this looks fine 👍

MartinGuindon:
Hello @asottile , the cleanup has been done and version 1.1 has been released. We should be good to go here!

asottile (approved):

PR #784, opened 2023-02-02
Add adamchainz/django-upgrade and move blacken-docs

adamchainz:
my new repository:

- [x] contains a license


asottile:
want to move `blacken-docs` next to yours while you're at it?

adamchainz:
Yup, done!

asottile (approved):

PR #799, opened 2023-03-10
Add hooks for LaTeX linters with valid repo name

joclement:
my new repository:

- [x] contains a license

Sorry for not having the name correctly in the first place.


asottile:
you checked the third box but don't adhere to it

joclement:
The `language` is `docker_image` and not `docker`. Are both not allowed?

PR #800, opened 2023-03-11
Add hook to check and strip GPS metadata

nuket:
Detect and remove GPS data from images using pre-commit hook. Usually happens by accident, but it does happen.

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -232,3 +232,4 @@
 - https://github.com/rubocop/rubocop
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
+- https://github.com/nuket/Eigendoxx


asottile:
hi -- your tool doesn't work unless `exiftool` is globally available which makes it not a great fit for the supported hooks page -- perhaps you could do something like `shellcheck-py` or write the tool in a language which can directly provide `exiftool` -- the template should really say `script` along with `system`

nuket:
Lol there's no way I'm reimplementing exiftool in another language, the complexity of parsing EXIF tags correctly is hilariously high. I tried declaring it as a `type: perl` project, but it seems to require build-essentials on the host system, which isn't much better of an option IMO. I'd argue for the use of a Docker container to solve all of this, but most devs don't seem to know how to use it effectively.

PR #805, opened 2023-03-22
Add python filename linter repo

ClementPinard:
Hi, thanks for this nice tool !

Added a small hook I developed because I was tired to see my colleagues using PascalCase for python filenames instead of snake_case.

my new repository:

- [ X ] is added to the bottom *or* with existing repos from the same account
- [ X ] contains a license
- [ X ] is not `language: system`, `language: script`, or `language: docker`
- [ X ] does not contain "pre-commit" in the name

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -233,3 +233,4 @@
 - https://github.com/dbt-checkpoint/dbt-checkpoint
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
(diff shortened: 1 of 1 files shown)


ClementPinard:
I see, makes sense if we just want to check that *.py are indeed in snake_case, thanks for the simple solution!

The way I see this repo is hopefully a more advanced tool in the future to rename your modules automatically (and all its imports), but I do agree that for now, it's not much more than a regex pattern that will fail if it's not empty.

Feel free to remove the PR if you think the tool is too premature.

PR #807, opened 2023-03-29
feat: added flutter-test

c-seeger:
my new repository:

- [x] contains a license
- [ ] is not `language: system`, `language: script`, or `language: docker`


c-seeger:
Looks like `language: flutter` does not exists, all other flutter hooks use `language: script`

c-seeger:
What is meant by `does not contain "pre-commit" in the name`, the name in the `.pre-commit-hooks.yaml` or the repository name? (or both?)

asottile:
the check mark is specifically about the repository name -- but it also wouldn't make sense in the hook id or name either

tests don't really make sense as part of pre-commit (they're far too slow and more likely to have your contributors turn off the whole thing) and the tool doesn't meet the portability standards to be on the hooks page

PR #808, opened 2023-03-31
added check-old-commit-date hook to all-repos.yaml

whosayn:
my new repository:

- [x] contains a license


asottile:
I'm not sure this is generic enough to be useful (what is "old"? why should one care?) -- or that `prepare-commit-msg` is the correct hook (usually that hook is for formatting the commit message rather than doing validation)

PR #809, opened 2023-03-31
Make pull request template more detailed

joclement:
This might accomplish less unfitting hooks being submitted via pull requests.


asottile:
I don't think this improves anything

PR #813, opened 2023-04-03
Add https://github.com/tobinjt/golang-coverage-check

tobinjt:
my new repository:

- [x] contains a license


asottile:
running tests as part of pre-commit is discouraged -- it's far too slow and will frustrate your contributors to turning off the tool entirely.  checks should generally be very quick static checks

PR #816, opened 2023-04-07
Add https://github.com/nrbeach/requirements_checker

nrbeach:
my new repository:

- [X] contains a license
- [X] is not `language: system`, `language: script`, or `language: docker`

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -234,3 +234,4 @@
 - https://github.com/dbt-checkpoint/dbt-checkpoint
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
+- https://github.com/nrbeach/requirements_checker


all-repos.yaml
```diff
@@ -234,3 +234,4 @@
 - https://github.com/dbt-checkpoint/dbt-checkpoint
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
+- https://github.com/nrbeach/requirements_checker
```
asottile, line 237: this doesn't really make sense as a pre-commit hook.  it doesn't matter whether a virtualenv is up to date when committing -- the virtualenv isn't part of the commit

nrbeach: The rationale behind the hook is primarily to prevent the user from committing code that has dependencies that aren't reflected in `requirements.txt`, i.e. adding code referencing new/updated modules.  Comparing the virtualenv's contents is admittedly a rather crude mechanism to achieve this. I wrote this primarily for my own use - to help keep `requirements.txt` in sync in the early stages of a project when I'm adding/removing modules frequently.

I completely understand if this is considered off-label use - please close the PR if this doesn't fit.

PR #819, opened 2023-04-13
Add pre-commit-buildifier

hofbi:
my new repository:

- [x] contains a license


hofbi:
What was wrong with this PR? I did not set the 'does not contain "pre-commit" in the name' but I just realized that you are probably talking about the name of the hook, not of the repo. The hook-id is just `buildifier`. So if there are no issues it would be great if you could reconsider this PR

hofbi:
but what is then the issue? I checked the existing hooks there are a bunch of hooks with pre-commit in the URL

asottile:
requirements change, feel free to research. I am not going to reiterate

the point of the checklist is so you can see you didn't meet the requirements and not even open the PR in the first place

PR #820, opened 2023-04-13
Add https://gitlab.com/brettops/hooks/verilog

bweir:
Hi, I removed `pre-commit` from the parent group name, so I am re-submitting for review. Thanks.

my new repository:

- [x] contains a license


asottile:
ah sorry, I thought it would be clear if `language: docker` is banned that `language: docker_image` would also be banned -- sorry about that

PR #822, opened 2023-04-18
Add django-ensure-makemessages to all-repos

nils-van-zuijlen:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -234,4 +234,4 @@
 - https://github.com/dbt-checkpoint/dbt-checkpoint
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
-- https://gitlab.com/nils-van-zuijlen/pre-commit-django-makemessages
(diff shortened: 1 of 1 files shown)


asottile:
this isn't a good fit for pre-commit as it requires the user to install things outside of the framework -- plus your `files` regex is wrong

nils-van-zuijlen:
I only found out about the naming requirements in the PR template, should I create an issue for having them in the "Contributing" section?

asottile:
no, the PR template is enough to show the requirement

nils-van-zuijlen:
Fair enough. Do you have an idea about how I can include the gettext dependency ? The other ones (python, manage.py file) should already be dealt with in the context of a django app.

asottile:
it's not those -- it's the "context of a django app" that's the problem -- you have no guarantee that the django app is set up properly and ready to go

nils-van-zuijlen:
Then I don't understand why https://github.com/jguttman94/pre-commit-gradle is in the list as it needs the gradle config to be set up and ready to go.

Also, you don't want to commit code that does not launch either way, no?

PR #835, opened 2023-06-02
add enforce-notebook-run-order

christopher-hacker:
my new repository:

- [x] contains a license


asottile:
- pre-commit is not capitalized (a few places you've spelled it "Pre-commit")
- your README example isn't valid yaml
- `name`, `entry`, and `types` are incorrect in your README example
- as far as I can tell the tool does not accept multiple filenames on the commandline so it won't work as-is with pre-commit

christopher-hacker:
@asottile - those changes are made and I believe this should be ready to merge now

asottile (approved):

PR #838, opened 2023-06-09
Add `mirrors-rome` to `all-repos.yaml`

JayOfferdahl:
my new repository:

- [X] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -234,3 +234,4 @@
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
 - https://github.com/christopher-hacker/enforce-notebook-run-order
+- https://github.com/JayOfferdahl/mirrors-rome


asottile:
the mirror repositories are meant as a stopgap and not a permanent solution -- if possible please propose the metadata to the tool itself instead

JayOfferdahl:
Ah -- appreciate the info. I'll submit a PR over there!

JayOfferdahl:
For any future travelers that find this: I ended up adding a local node repository to my configuration as a temporary solution:
```
    - repo: local
      hooks:
          - id: rome
            entry: rome
            name: rome
            args: [check]
            language: node
            types_or: [javascript, ts, jsx, tsx, json]
            additional_dependencies:
                - rome
```

I'm trying to get the metadata submitted to the `rome/tools` project but am running into a few snags with `pre-commit` initializing the repo because Cargo is unable to install the crate -- something about finding a virtual manifest instead of a package manifest. I'll need to dig a bit more on this.

PR #839, opened 2023-06-12
Add `kxue43/post-merge-hooks` to "Supported hooks"

kxue43:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -234,3 +234,4 @@
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
 - https://github.com/christopher-hacker/enforce-notebook-run-order
+- https://github.com/kxue43/post-merge-hooks


kxue43:
@asottile Sorry I'm not very familiar with `deploy.yml`. Could you give me some idea why it failed? Does it have anything to do with my setting `verbose: true` in my hooks?

asottile:
yes, that is not allowed

but also:
- your repo's name is too generic, it doesn't help someone understand what it does
- I don't think your hooks provide anything on top of what `pre-commit` already does -- in fact the file filtering seems to miss a main point of pre-commit in that you don't need to reinvent file filtering (`files` / `types` does that already).  in other words, someone could write a `repo: local` hook that works as well with zero external repos

kxue43:
> yes, that is not allowed
> 
> but also:
> 
> * your repo's name is too generic, it doesn't help someone understand what it does
> * I don't think your hooks provide anything on top of what `pre-commit` already does -- in fact the file filtering seems to miss a main point of pre-commit in that you don't need to reinvent file filtering (`files` / `types` does that already).  in other words, someone could write a `repo: local` hook that works as well with zero external repos

I can change the repo name.

As to file filtering, I'm sure it does so for "pre-commit" hooks, but does it pass them for "post-merge" hooks as well? On my repo I created a tag `test2` where `pass_filenames` is set to `true`, and I put `print(args.filenames)` in the code. When the `git pull` does change some files, `args.filenames` is shown to be `[]`. Moreover, my code filters the files changed in a `git pull` _for file changes that user care about_, and only run shell command on these changes, not on any changes.

asottile:
yeah so that's where `pass_filenames: false` comes into play

kxue43:
> yeah so that's where `pass_filenames: false` comes into play

I'm saying I've changed `pass_filenames` to the default value of `true`, and it's still not passing changed file names for post-merge hooks.

asottile:
`post-merge` does not execute on files

kxue43:
So my hook does do something that is not possible with a `repo: local` hook right now.

In fact, I created a local hook as below:

```yaml
repos:
  - repo: local
    hooks:
      - id: test-hook
        name: test-hook
        language: system
        entry: echo hello
        files: ^testfile$
        stages: [pre-commit, post-merge]
        verbose: true
```

On pre-commit changes to `testfile`, the pre-commit hook does run and print out "hello testfile". When a `git pull` change `testfile`, the post-merge hook is skipped because "no files to check". I don't think it's accurate to say `pre-commit` already provides everything my post-merge hooks do.

PR #843, opened 2023-06-22
Adding `bump-helm-chart-version` to supported hooks list

devqik:
`bump-helm-chart-version` is a pre-commit hook written in Python 3. It is used to automatically bump the Helm chart patch version if the chart has changes in the /templates folder or changes in helper functions files.

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -234,3 +234,4 @@
 - https://gitlab.com/engmark/vcard
 - https://github.com/Data-Liberation-Front/csvlint.rb
 - https://github.com/christopher-hacker/enforce-notebook-run-order
+- https://github.com/devqik/bump-helm-chart-version


asottile:
I don't think this hook is a good idea.  version bumps should be a conscious decision by the developer.  auto-bumping a patch version always cannot be correct.

unrelated code feedback:

- this is a bad idea and will not work anyway: https://github.com/devqik/bump-helm-chart-version/blob/25c09ff09d784fa29671b39460d519d4470572dc/bump_helm_chart_version/bump_chart_version.py#L12-L16
- hooks should not touch the staging area: https://github.com/devqik/bump-helm-chart-version/blob/25c09ff09d784fa29671b39460d519d4470572dc/bump_helm_chart_version/bump_chart_version.py#L99

devqik:
Thanks for the feed back @asottile .
Was a challenge I have seen in my position that me and some colleagues don't remember to bump the patch version and that causes the usage of an old helm chart.

Was useful in my case to automate the process for small changes (If there is a major or minor version we have to do it manually)
Thanks again for the feedback 🙌 Cheers.

PR #847, opened 2023-07-06
Add terraform docs repo

william-richard:
my new repository:

- [X] contains a license
- [ ] is not `language: system`, `language: script`, `language: docker`, or `language: docker_image`
  - It contains 3 hooks, each with a different language. `golang`, `system` and `docker`


asottile:
you've also got #754 -- can you rebase that one instead?

PR #851, opened 2023-07-14
sections/advanced.md: Avoid warning on old action

t-b:
Old actions which use node 12 currently spew a warning, see [1].

Let's use the latest version in the documentation here to help future users.

[1]: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


asottile:
this isn't really intended to be kept up to date -- but fine

PR #853, opened 2023-07-16
Add `pre-commit/sync-pre-commit-deps`

mxr:
my new repository:

- [x] contains a license
  ^-- `pre-commit` is in the name since this is a hook that modifies `pre-commit` config

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -12,6 +12,7 @@
 - https://github.com/pre-commit/mirrors-scss-lint
 - https://github.com/pre-commit/mirrors-yapf
 - https://github.com/pre-commit/pygrep-hooks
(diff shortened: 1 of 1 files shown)


asottile (approved):
has conflicts though -- maybe could move it to `pre-commit/` too if you'd want? if we want to make this more official?

mxr:
Sure more than happy to move it there. Thanks!

mxr:
Seems like I don't have access to initiate the transfer into that org, can you grant access temporarily?

mxr:
OK, initiated. Thanks!

asottile:
done!

asottile (approved):

PR #859, opened 2023-08-15
Add mrsimonemms/pre-commit-hooks repo

mrsimonemms:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -237,3 +237,4 @@
 - https://github.com/christopher-hacker/enforce-notebook-run-order
 - https://github.com/mxr/unkey
 - https://github.com/Mateusz-Grzelinski/actionlint-py
+- https://github.com/mrsimonemms/pre-commit-hooks


mrsimonemms:
@asottile how come this was closed without comment - what's wrong?

asottile:
image

asottile:
and even that's not correct as you haven't satisfied the third requirement either.  why even bother with a checklist if it's being blatantly ignored

mrsimonemms:
How come this is enforced when there are many existing repos which are variations on "pre-commits-hooks" already in there?

It would also be nice to have a comment explaining why - it's not a great look for a project to just close community contributions so tersely.

mrsimonemms:
Ok, you're clearly in a grump. I'm going to write this off as a bad job.

Goodbye. I hope the rest of the day puts you in a better mood

asottile:
because the list is nearly 10 years old and has grown and changed over time

PR #878, opened 2023-09-24
add Astyle hook (igrr/astyle_py) to all-repos.yaml

igrr:
my new repository:

- [x] contains a license


all-repos.yaml
```diff
@@ -238,3 +238,4 @@
 - https://github.com/mxr/unkey
 - https://github.com/Mateusz-Grzelinski/actionlint-py
 - https://github.com/andrewring/github-distributed-owners
+- https://github.com/igrr/astyle_py
```
asottile, line 241: a few small things:

- `pass_filenames` defaults to `true` so you shouldn't specify that
- `install_requires` is currently `==` pinning everything which is inflexible for users -- this should probably use `>=` instead

igrr: Thanks, will fix that!

I will probably keep wasmtime pinned, though, since nearly every release has some breaking changes at this point.

igrr: @asottile Have made both changes and tagged the new release.

PR #882, opened 2023-10-09
Update deprecated `yapf` to the official source in `all-repos.yaml`

cerob:
Update yapf

my new repository:

- [x] contains a license


asottile:
your commit message is bad

cerob:
As the `mirror-yapf` repository states it is deprecated, it is time to use the official repository. An explanation is also here: https://github.com/pre-commit/mirrors-yapf

asottile:
"Update all-repos.yaml" tells nothing -- explain what you're doing and why *in the commit message*

PR #890, opened 2023-10-26
feat: Add dc-avro lint pre-commit to all-repos

svdimchenko:
Add dc-avro lint pre-commit

my new repository:

- [x] contains a license

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,11 @@
+# Generated artifacts
 *.html
 *.pyc
 *.swp
 /all-hooks.json
 /build
 /node_modules
+
+# IDE
+.idea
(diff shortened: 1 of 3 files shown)


asottile:
it looked like a mistake?

asottile:
no your pr looks like a mistake because it has a bunch of unrelated changes

svdimchenko:
> no your pr looks like a mistake because it has a bunch of unrelated changes

okay, you can keep it closed then. However I think `.gitignore` contains bunch of misconfiguration then 😄

asottile:
no. my gitignore does not need your junk for tools I don't use

PR #891, opened 2023-10-27
feat: Add dc-avro lint

svdimchenko:
Add dc-avro to lint avro schemas in pre-commit

- [x] contains a license


asottile:
your regex in your `.pre-commit-hooks.yaml` is not correct -- it would disallow files with whitespace in them

svdimchenko:
Current regex `\S+\.avsc$` does not allow whitespace and that's OK for avro schema files. Could you provide and example when do you need to keep whitespace ?

asottile:
`my schema.avsc`

svdimchenko:
ah, I've never met such naming for avro schemas. So feel free to open an issue in corresponding repo.

PR #892, opened 2023-10-28
Add check-comma-in-class-attributes repo to all-repos.yaml file

marcosdotme:
my new repository:

- [x] contains a license


asottile:
I believe flake8-commas accomplishes the same thing but is a little more battle tested

PR #893, opened 2023-11-03
Docs: add note about pre-push hooks not being triggered on ref deletions

whophil:
Documents behavior described in https://github.com/pre-commit/pre-commit/issues/3050

### DIFF
--- a/sections/advanced.md
+++ b/sections/advanced.md
@@ -182,6 +182,8 @@ Note that you need to be using at least git 2.24 for this hook.
 
 `pre-push` is triggered on `git push`.
 
+note: `pre-push` hooks are not triggered for ref deletions.
+
 environment variables:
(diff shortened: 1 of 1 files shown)


sections/advanced.md
```diff
@@ -182,6 +182,8 @@ Note that you need to be using at least git 2.24 for this hook.
 
 `pre-push` is triggered on `git push`.
 
+note: `pre-push` hooks are not triggered for ref deletions.
+
```
asottile, line 186: there's a bit more nuance to this than just that -- the variables below imply this is already the case since there must be a from-ref and a to-ref

whophil: Based on the environment variables, I was expecting `PRE_COMMIT_TO_REF` to be `0000000000000000000000000000000000000000` here.

Would you propose an alternate wording, or not documenting this at all?

PR #899, opened 2023-11-15
docs: add new pre-commit hook for conventional commits to all-repos.yaml

TribuneX:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -240,3 +240,4 @@
 - https://github.com/andrewring/github-distributed-owners
 - https://github.com/Mermeid-Designs/pydantic-hooks
 - https://github.com/igrr/astyle_py
+- https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit


all-repos.yaml
```diff
@@ -240,3 +240,4 @@
 - https://github.com/andrewring/github-distributed-owners
 - https://github.com/Mermeid-Designs/pydantic-hooks
 - https://github.com/igrr/astyle_py
+- https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit
```
asottile, line 243: the `name` is a bit too long -- this is going to cause pre-commit to make its output very wide

also `always_run: true` is not needed since this is a `prepare-commit-msg` hook it will always receive the commit message filename

the executable `hook-entrypoint` also isn't descriptive of what you're doing -- I'd recommend naming it in a way that identifies your project

TribuneX: Thanks for the feedback.
I added a link to a new shorter repo (to avoid breaking the hook for existing users) and adapted the hook config based on your feedback.

asottile: oh not the repo name -- this `name`: https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit/-/blob/main/.pre-commit-hooks.yaml?ref_type=heads#L2

TribuneX: I see! I changed it to a simpler and shorter [version](https://gitlab.com/codementors/pre-commit/conventional-commits/-/blob/main/.pre-commit-hooks.yaml#L2). However, I also like the new repo name better, I am okay with using the new one to be listed on pre-commit.com

asottile: I think the new repo name is worse at communicating what your repo does

TribuneX: Understood. I added the requested changes to the [original repository](https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit/-/commit/d6ef5dd36bb4b1e69b1306955fa18055bc08a688). I removed the last commit from this PR.

asottile (approved):

PR #904, opened 2023-11-26
Add github.com/scop/wrun

scop:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -206,6 +206,7 @@
 - https://github.com/adamchainz/django-upgrade
 - https://github.com/scop/pre-commit-shfmt
 - https://github.com/scop/pre-commit-perlcritic
+- https://github.com/scop/wrun
 - https://github.com/BlankSpruce/gersemi
 - https://github.com/realm/SwiftLint
 - https://gitlab.com/bmares/check-json5


all-repos.yaml
```diff
@@ -206,6 +206,7 @@
 - https://github.com/adamchainz/django-upgrade
 - https://github.com/scop/pre-commit-shfmt
 - https://github.com/scop/pre-commit-perlcritic
+- https://github.com/scop/wrun
```
asottile, line 209: this downloads tools at runtime rather than at build time so it's not a good fit for the framework -- perhaps you can chime in on https://github.com/pre-commit/pre-commit/issues/1453 and https://github.com/pre-commit/pre-commit/pull/3020 which are trying to solve this more generically (and correctly)

asottile: the problem is more that there's no write protection to the pre-commit cache at runtime (intentionally) and so writes in general break the cache integrity and at worst concurrent writes permanently poison the cache and so writing at runtime is unsupported

scop: A workaround for that could be if pre-commit.ci provided a way to cache additional paths, not just the pre-commit cache.  Whatever happens in them would be the responsibility of something else besides pre-commit. Can imagine that this would likely make sharing such additional caches pretty much a no go, as well as their retention suboptimal.

asottile: it's not a pre-commit.ci issue -- it's a pre-commit issue.  writes to pre-commit's cache at runtime are unsafe

scop: > I'll likely remove the helper option to (ab)use a subdir in pre-commit's cache. 

FTR, this option has been [removed since](https://github.com/scop/wrun/commit/dd08280265481545a6563835a6317008597b97b5), wrun no longer helps or recommends the user in any way to touch pre-commit's cache.

So in that sense I suppose the blocker against adding this in pre-commit's supported hooks list is now gone, and the addition could be reconsidered. My branch is still there.

PR #908, opened 2023-12-02
Add gitlab-ci-tools and codeowners-generator

c0m1c5an5:
Add gitlab-ci-tools repo that provides a shellcheck, linter and formatter tools for gitlab ci pipeline files.
Add codeowners-generator repo that provides a hook that keeps CODEOWNERS up to date. 

my new repository:

- [X] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -242,3 +242,5 @@
 - https://github.com/igrr/astyle_py
 - https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit
 - https://github.com/numpy/numpydoc
+- https://github.com/c0m1c5an5/gitlab-ci-tools
+- https://github.com/c0m1c5an5/codeowners-generator


all-repos.yaml
```diff
@@ -242,3 +242,5 @@
 - https://github.com/igrr/astyle_py
 - https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit
 - https://github.com/numpy/numpydoc
+- https://github.com/c0m1c5an5/gitlab-ci-tools
+- https://github.com/c0m1c5an5/codeowners-generator
```
asottile, line 246: `require_serial` does not guarantee that all files are passed in a single invocation so I don't think this hook is correct

all-repos.yaml
```diff
@@ -242,3 +242,5 @@
 - https://github.com/igrr/astyle_py
 - https://gitlab.com/codementors/pre-commit/add-issue-number-to-conventional-commit
 - https://github.com/numpy/numpydoc
+- https://github.com/c0m1c5an5/gitlab-ci-tools
```
asottile, line 245: this could probably depend on `shellcheck-py` and then you wouldn't require manual setup by the user to have shellcheck available which makes at least one of the hooks here not up to standards (as it's essentially a `language: system` hook)

c0m1c5an5: Yes, thank you for the suggestion. `shellcheck-py` fits perfectly. Yet other hooks (`gitlab-ci-fmt` and `gitlab-ci-lint`) also rely on local system utilities, so I am in a process of rewriting them to fit this requirement.

PR #912, opened 2023-12-19
Update new-hooks.md - make sentence clearer

noamraph:
I changed "The hook must exit nonzero on failure or modify files" into "The hook must exit nonzero on failure or if it modifies files". The old sentence can be read that the hook must either exit nonzero on failure, or modify files. It doesn't make much sense, but anyway I think it's confusing.

### DIFF
--- a/sections/new-hooks.md
+++ b/sections/new-hooks.md
@@ -4,7 +4,7 @@ installable package (gem, npm, pypi, etc.) or exposes an executable, it can be
 used with pre-commit. Each git repo can support as many languages/hooks as you
 want.
 
-The hook must exit nonzero on failure or modify files.
+The hook must exit nonzero on failure or if it modifies files.
 
 A git repo containing pre-commit plugins must contain a `.pre-commit-hooks.yaml`
 file that tells pre-commit:


asottile:
the old wording is correct.  it's either or

noamraph:
@asottile I think I'm missing something here. If the hook changes files, shouldn't it return a nonzero code?

I'm trying to fix this bug: https://github.com/python-poetry/poetry/issues/8512. I want to use the poetry-export hook to make sure my requirements.txt file is up to date with the lock file. Currently, running the hook updates the file, but always returns 0. This means that pre-commit returns 0 (success). This means that if I run pre-commit in a CI pipeline, it won't detect if a user didn't use pre-commit so the requirements.txt file is not correct.

As an example, see the [whitespace fixer](https://github.com/pre-commit/pre-commit-hooks/blob/42bc79146f8b566f640ff773057d48a9416763a7/pre_commit_hooks/trailing_whitespace_fixer.py#L98 ) - it returns 1 if it fixed a file.

asottile:
have you tried it? that issue is incorrect

noamraph:
@asottile yes, I ran pre-commit, and saw that it returned with exit code 0, even though the requirements.txt file was missing or wasn't up to date. (It was updated by the hook, but pre-commit returned 0 instead of 1)

asottile:
that's not how pre-commit works -- if a source file is modified that counts as a failure

asottile:
tiny demo:

```console
$ touch f
$ git add f
$ cat > .pre-commit-config.yaml <<EOF
> repos:
> -   repo: local
>     hooks:
>     -   id: modifies
>         name: modifies
>         entry: bash -c 'echo hi > f; exit 0'
>         pass_filenames: false
>         always_run: true
>         language: system
> EOF
$ git add .pre-commit-config.yaml 
$ pre-commit; echo $?
modifies.................................................................Failed
- hook id: modifies
- files were modified by this hook
1
```

noamraph:
@asottile Indeed! I think that the problem was that I didn't add the requirements.txt file to git before testing the behavior of pre-commit.

I want to still suggest another wording, that perhaps would have helped me realize my mistake sooner. What about:

> On failure (meaning file contents are not as expected), the hook must either exit with a nonzero code, or modify the files. Pre-commit checks if any committed file was changed by the hook, which lets it report a failure even if the hook returns 0.

(I added the description of what a failure means because it may also mean just an exception in the hook.)

WDYT?

asottile:
no.  it is already concise and accurate

PR #914, opened 2023-12-23
Support gn-format hook in all-repos.yaml

KindDragon:
my new repository:

- [x] contains a license

This hook for formatting GN build files https://gn.googlesource.com/gn/

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -244,3 +244,4 @@
 - https://github.com/numpy/numpydoc
 - https://github.com/hakancelikdev/unimport
 - https://github.com/hakancelikdev/unexport
+- https://github.com/KindDragon/gn-build-py


all-repos.yaml
```diff
@@ -244,3 +244,4 @@
 - https://github.com/numpy/numpydoc
 - https://github.com/hakancelikdev/unimport
 - https://github.com/hakancelikdev/unexport
+- https://github.com/KindDragon/gn-build-py
```
asottile, line 247: just a few things:
- `types: [file]` is the default so you can leave that out: https://github.com/KindDragon/gn-build-py/blob/d85d1eded700c76a912a7ca38f8d4963dbb995e4/.pre-commit-hooks.yaml#L6C3-L6C16
- `rev: latest` (1) does not work and (2) [is not supported](https://pre-commit.com/#using-the-latest-version-for-a-repository): https://github.com/KindDragon/gn-build-py/blame/d85d1eded700c76a912a7ca38f8d4963dbb995e4/README.md#L11

KindDragon: I've made corrections based on your suggestions

asottile: `rev: ''` is also similarly not supported

KindDragon: Yes, but I added to README.md to call `pre-commit autoupdate` after that to use latest available revision

PR #916, opened 2023-12-28
docs: adds python-poetry

yxtay:
python-poetry/poetry:

- [x] contains a license

astral-sh/ruff-pre-commit:

- [x] contains a license


asottile:
please split your patch into two separate PRs (and search for duplicates) and keep existing things in the same spot if renamed

PR #917, opened 2023-12-28
docs: adds python-poetry

yxtay:
[python-poetry/poetry](https://github.com/python-poetry/poetry):

- [x] contains a license


asottile:
please search for duplicates next time like I asked!

PR #918, opened 2023-12-28
docs: updates ruff pre-commit

yxtay:
Updates charliermarsh/ruff-pre-commit to [astral-sh/ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit):

- [ ] contains a license
- [ ] is not `language: system`, `language: script`, `language: docker`, or `language: docker_image`


yxtay:
> does not contain "pre-commit" in the name

I'm not sure what this mean exactly. The repo name contains "pre-commit". However, given that this is updating a previously accepted repo, I suppose it is ok.

PR #923, opened 2024-01-12
Describe how `identity` hook can be used to print messages

namurphy:
This PR mentions that the `identity` hook can be used to print a message when `pre-commit` is run.

For a bit of context, a while back I was looking for a way to print out a message with a link to our troubleshooting guide when a new contributor encountered a `pre-commit` failure.  I couldn't find anything in the documentation, so I ended up finding https://github.com/pre-commit/pre-commit/issues/1565.  It took a few iterations before settling on the very helpful suggestions to use the `identity` hook.  I wanted to add this to the documentation since it would have probably saved me ∼2–3 hours of time and prevented a lot of frustration.  

Thank you!

### DIFF
--- a/sections/advanced.md
+++ b/sections/advanced.md
@@ -357,7 +357,8 @@ The currently available `meta` hooks:
 =r=
     =c= [`identity`](_#meta-identity)
     =c= a simple hook which prints all arguments passed to it, useful for
-        debugging.
+        debugging or printing a help message (if `pass_filenames` is set to
+        `false`).
 ```
 
 ## automatically enabling pre-commit on repositories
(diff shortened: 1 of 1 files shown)


asottile:
above this is already:

>     a simple hook which prints all arguments passed to it

which is sufficient already

asottile:
fwiw I don't recommend what you're doing either. you've added noise to every one of your runs and I don't want the docs to encourage that either

namurphy:
Fair enough.  Primarily I was looking for a way to have a link to the troubleshooting guide to get printed out only when there's a test failure, and only in CI.  I'll move this question to stackexchange perhaps, since it's out of scope of this PR.

PR #924, opened 2024-01-17
Add exif-stripper to all-repos.yaml

stefmolin:
my new repository:

- [X] contains a license


all-repos.yaml
```diff
@@ -245,3 +245,4 @@
 - https://github.com/hakancelikdev/unimport
 - https://github.com/hakancelikdev/unexport
 - https://github.com/KindDragon/gn-build-py
+- https://github.com/stefmolin/exif-stripper
```
asottile, line 248: `require_serial: false` is the default so you can leave that out!

asottile (approved):

Issue #797, opened 2023-03-07
It’s not clear which languages can’t be used for local hooks


Jayman2000:
According [the repository local hooks section](https://pre-commit.com/#repository-local-hooks),

> local hooks can use any language which supports [`additional_dependencies`](https://pre-commit.com/#config-additional_dependencies) or `docker_image` / `fail` / `pygrep` / `script` / `system`.

It’s not clear to me which languages can’t be used. For example, the descriptions for both [`python`](https://pre-commit.com/#python) and [`ruby`](https://pre-commit.com/#ruby) don’t explicitly mention that those languages support `additional_dependencies`, but they do indeed support `additional_dependencies`. This makes me question whether or not you can create (for example) a local `dotnet` hook. [The description for `dotnet`](https://pre-commit.com/#dotnet) doesn’t explicitly say that `additional_dependencies` is supported, but it also doesn’t say that `additional_dependencies` isn’t supported.

flying-sheep:
Similarly, I can’t find documentation about what languages are supported on pre-commit.ci. E.g. `docker_image` isn’t, what else?

asottile:
just the docker based ones, everything else should work

soxofaan:
related: 
- #980

PR #927, opened 2024-01-29
Extend golang docs: additional dependencies and tooling semantics

lubgr:
Minor improvements to the docs on `golang` hooks:

- In `golang` hooks, `additional_dependencies` can be used. According to the existing docs on other language hooks, I have added a note on that. See also issue #797.
- The second paragraph is a summary of my findings from setting up `golang` hooks. It might be worth adding this for the benefit of others?

### DIFF
--- a/sections/new-hooks.md
+++ b/sections/new-hooks.md
@@ -314,6 +314,9 @@ The hook repository must contain go source code.  It will be installed via
 and the [`entry`](#hooks-entry) should match an executable which will get installed into the
 `GOPATH`'s `bin` directory.
 
+This language supports `additional_dependencies` and will pass any of the values directly to `go
+install`. It can be used as a `repo: local` hook.
+
 _changed in 2.17.0_: previously `go get ./...` was used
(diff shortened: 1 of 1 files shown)


asottile:
I would remove the second paragraph.  there are many poorly written tools in every language so it is not necessary to call them out in the golang section

lubgr:
Thanks for having a look so quickly!

The second paragraph was not commenting on UI of Go tools, but pointing out a fundamental mismatch of how the entire Go ecosystem works (including the compiler), where individual files have no meaning, only packages contained in directories have. The tools that refuse to work on individual files are hence fully streamlined and provide the least surprise for Go devs. As a consequence, it's hard to write Go hooks for Go code that work as nicely as hooks written in other languages.

But thinking about it again, this only applies to Go hooks written to work with _Go code_, not Go hooks generally, for any other language.

So long story short, I have remove the paragraph 🙂

asottile (approved):

PR #929, opened 2024-02-01
add flake8-staged-diff in all-repos.yaml

bagerard:
my new repository:

- [x] contains a license


asottile:
given what you've already quoted me on in the readme, I don't want to encourage this

PR #932, opened 2024-02-17
Added clean-dotenv to all-repos.yaml

hija:
Hi there,

I created a pre-commit hook which creates a .env.example file (derived from the .env file).

My new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -246,3 +246,4 @@
 - https://github.com/stefmolin/exif-stripper
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
+- https://github.com/hija/clean-dotenv


hija:
Yikes. I forgot to set the repo to public. Fixed it now but don't know how to trigger a rerun of the failed workflow. 😞

all-repos.yaml
```diff
@@ -247,3 +247,4 @@
 - https://github.com/stefmolin/exif-stripper
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
+- https://github.com/hija/clean-dotenv/
```
asottile, line 250: you don't need `types`, `exclude_types`, or `exclude` in your hook definition

`always_run: true` is ~usually wrong -- but in this case is probably fine because it's intentionally reacting to a file which is not part of the repository

a few other things:
- I would make the env directory required and not recurse (this is going to be extremely slow on large repositories)
- I would not support case insensitive `.env` -- there's only one spelling of env so no reason to support `.eNv` or other permutations

asottile: as for env directory -- I mean more that there's a single place you would put the `.env` file -- recursing is not a good idea here.

you can always specify the hook more than once if they have multiple env files though that seems unlikely.  you can probably default this directory to `.`

hija: Ah! Now I got you. Thank you for elaborating further.

> you can always specify the hook more than once if they have multiple env files though that seems unlikely.

Did not think about that. Good point. Removed the recursion!

So we should be clear now. Thanks for the feedback 👍

asottile (approved):

PR #937, opened 2024-02-27
Add enforce-https to all-repos.yaml

GideonBear:
my new repository:

- [x] contains a license


asottile:
sorry, can't accept GPL

also I suspect this entire repo can be replaced easily with `pygrep`

GideonBear:
> sorry, can't accept GPL

Out of curiosity, is that because of legal reasons, or because you're against it on principle?

asottile:
both.  I can't review your code for risk of making a derivative work and gpl is a cancer

PR #939, opened 2024-03-04
Add bufbuild/buf to all-repos.yaml

wwuck:
https://github.com/bufbuild/buf has included some pre-commit hooks

my new repository:

- [x] contains a license


asottile:
`pass_filenames: false` is usually wrong and results in always running everything always

asottile:

PR #949, opened 2024-04-06
Add xenon

davetapley:
https://github.com/rubik/xenon/pull/36

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -247,3 +247,4 @@
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
 - https://github.com/hija/clean-dotenv
+- https://github.com/rubik/xenon


all-repos.yaml
```diff
@@ -247,3 +247,4 @@
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
 - https://github.com/hija/clean-dotenv
+- https://github.com/rubik/xenon
```
asottile, line 250: I believe this should be `types: [python]` here: https://github.com/rubik/xenon/blob/9d06d21515d9956d4c270aa04b05d4f1751f9516/.pre-commit-hooks.yaml#L7

davetapley: Waiting on https://github.com/rubik/xenon/pull/54

asottile:

PR #952, opened 2024-04-09
Add toodaloo hooks to repos

mrsimonemms:
my new repository:

- [x] contains a license

Add support for [Toodaloo](https://github.com/mrsimonemms/toodaloo) to track TODOs in your source code


asottile:
`pass_filenames: false` is not a good fit for the framework because it necessarily means you're running on all files all the time

mrsimonemms:
@asottile yes, that's by design. I need to generate a definitive list of TODOs at each commit rather than an incremental list of new things.

If there's a better way of achieving this, I'd love to know how it can be done.

PR #955, opened 2024-04-17
Add the link to check-poetry

derkweijers:
my new repository:

- [x] contains a license


asottile:
this has previously been rejected -- please search previous pull requests.  also it doesn't need to exist -- you can override `args` / `entry` in the local pre-commit configuration

PR #956, opened 2024-04-22
Add `if-changed` to all-repos.yaml

jun-sheaf:
`if-changed` is a pre-commit that detects changes in `if-changed`-fenced code. It then alerts the developer to make changes in related files, if it hasn't been done. This is a general tool used to update enums, implementations, messages, and various other things that may be related.

my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -247,3 +247,4 @@
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
 - https://github.com/hija/clean-dotenv
+- https://github.com/mathematic-inc/if-changed


jun-sheaf:
> * I'm not sure this tool works correctly given lots of files or when run in parallel with pre-commit -- does it require all files in one pass?

`if-changed` only checks files changed between two revisions.

> this doesn't feel like a good fit for pre-commit -- if you have "synchronized" code structures you should have tests for them -- not bypassable linters

In a perfect world, sure. But in practice, testing synchronized code is a type of "meta-programming" that ends up becoming a maintenance burden and often requires a lot of development to get correct. If it eases your mind, several companies have a version of this pre-commit (e.g. the big G) in their orgs which are used heavily.

asottile:
> if-changed only checks files changed between two revisions.

how does it do that? I don't see that possibly being hooked up in the configuration that's in the repo

> If it eases your mind, several companies have a version of this pre-commit (e.g. the big G) in their orgs which are used heavily.

if anything that's a minus, google's development practices are generally bad for the average software shop

jun-sheaf:
> how does it do that? I don't see that possibly being hooked up in the configuration that's in the repo

See https://github.com/mathematic-inc/if-changed/blob/main/bin/if-changed.rs

> if anything that's a minus, google's development practices are generally bad for the average software shop

I'm really not sure where this comes from. Empirically, Google has a few _billion_ lines of code that is only workable because of all the development infrastructure they put into it. The only reason I can think that Google's development practice is "bad" is because there is a high focus on automation and the "average software shop" doesn't usually have capacity (or experience) to automate and maintain everything.

Also, rather than claiming a specific practice is "bad", I'd appreciate something more constructive like "there are better options such as so and so".

In any case, IIUC, this list is supposed to help others discover plugins that may help their situation. I don't really understand why we are doing a code/product review over a utility that others _are_ using. If the list is a "curated" list, it'd be great if you can specify this because FWIU this is a community-driven exhaustive list.

asottile:
then at the very least you're missing `stages` in your config

Google's development practices work well when you have many thousands of developers but are inefficient for smaller orgs (read: anyone that's not Google)

jun-sheaf:
> then at the very least you're missing stages in your config

Done. Do you want me to make a release as well?

asottile:
I don't think this works for the `pre-commit` stage from what I've read in the code -- only `pre-push` ?

jun-sheaf:
This will compare staged/untracked changes against head by default.

jun-sheaf:
@asottile Gentle ping. Please close this PR if you do not feel it fits the list.

asottile:

PR #957, opened 2024-04-24
add new mirror of prettier

kbroch-rivosinc:
my new repository:

- [ x] is added to the bottom *or* with existing repos from the same account
- [ x] contains a license
- [ x] is not `language: system`, `language: script`, `language: docker`, or `language: docker_image`
- [ x] does not contain "pre-commit" in the name

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -247,3 +247,4 @@
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
 - https://github.com/hija/clean-dotenv
+- https://github.com/rbubley/mirrors-prettier


asottile:
no thanks

kbroch-rivosinc:
This is unfortunate.  
I really ❤️ `pre-commit` for the effort the maintainers make to support multiple languages.
I like `prettier` as it supports the formatting/linting of many languages.

I thought @rbubley solution was good to manage the problem of versioning between prettier and pre-commit and to take the burden off of you having to maintain it.

Please let me know if there's anything else that can be done technically to have you endorse this solution  on pre-commit.com.

PR #960, opened 2024-05-05
Added `mirrors-pytest` to `all-repos.yaml`

Repiteo:
my new repository:

- [x] contains a license

### DIFF
--- a/all-repos.yaml
+++ b/all-repos.yaml
@@ -247,3 +247,4 @@
 - https://github.com/hhatto/autopep8
 - https://github.com/rhysd/actionlint
 - https://github.com/hija/clean-dotenv
+- https://github.com/repiteo/mirrors-pytest


asottile:
pytest should not be run with pre-commit

Repiteo:
Damn, couldn't have chosen a more qualified person to shoot this idea down lmao

asottile:
there's a lot of writing about this amongst the pre-commit *and* pytest issue tracker(s) -- here's a short summary: https://github.com/pre-commit/pre-commit-hooks/issues/291#issuecomment-394167917

Repiteo:
D'oh! I only checked the PRs for `pytest`, somehow didn't think to check issues as well. At least my specific use-case is rather small so I can get away with using it for the time being, but I can certainly see why you wouldn't want to "endorse" this practice by adding it officially.

PR #968, opened 2024-05-29
LICAdd sql lint hook

RobertLD:
my new repository:

- [x] contains a license


RobertLD:
Funny! Alright then, easy enough. Thank you

PR #971, opened 2024-06-09
Add SQLFluff hook wrapper for linting SQL embedded in python strings

RobertLD:
my new repository:

- [x] contains a license

This hook is similar to the base SQLFluff hook, but utilizes the python AST to lint/fix SQL embedded within string objects in python.


asottile:

PR #974, opened 2024-06-14
Clarify install for those unfamiliar with Python

Flimm:
- Clarify that pre-commit is the name of this tool, it is not just the name of the Git hook. This point is not obvious to newcomers. Clarify this at the top.
- Clarify that pre-commit is targetted to all programmers, not just Python programmers, even though it does require Python to run.
- Not all computers come with Python preinstalled, so mention this requirement.
- Start with the simplest installation method, which is pipx. Those who are not familiar with pipx can quickly skip to the next installation method.
- Allow the .pyz installation method to take up the most space, as it should be the one that is easiest for those who aren't familiar with Python or zipapps.
- Move the `pip install` method to last, as it the least easy to understand for non-beginners. On some systems, it will fail completely if not in a virtual environment.
- Delete the first section of the quick start, as it is duplicate information.

### DIFF
--- a/sections/install.md
+++ b/sections/install.md
@@ -1,39 +1,57 @@
-Before you can run hooks, you need to have the pre-commit package manager
-installed.
-
-Using pip:
+pre-commit is a command-line tool. To check to see if pre-commit is installed,
+run it like this:
 
 ```bash
-pip install pre-commit
+pre-commit --version
 ```
 
-In a python project, add the following to your requirements.txt (or
-requirements-dev.txt):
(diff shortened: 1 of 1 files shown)


asottile:
no thanks

asottile:
I can't imagine a world where you would think a drive by rewrite of the docs without an issue or discussion would be accepted

Flimm:
OK. If that's what you prefer, I'll do it your way. I'll submit issues first. Thank you for your response, that's helpful.

PR #979, opened 2024-06-24
Add some cross-references related to repo-local hooks

soxofaan:


sections/plugins.md
```diff
@@ -72,7 +72,8 @@ from.
 ```table
 =r=
     =c= [`repo`](_#repos-repo)
-    =c= the repository url to `git clone` from
+    =c= the repository url to `git clone` from,
+        or the special sentinel value [`local`](#repository-local-hooks).
```
asottile, line 76: it can also be `meta` as well

soxofaan: :+1:
added reference to that as well

asottile (approved):

PR #980, opened 2024-06-24
Document that additional_dependencies is supported for python

soxofaan:


sections/new-hooks.md
```diff
@@ -384,6 +384,10 @@ The hook repository must be installable via `pip install .` (usually by either
 executable that will match the [`entry`](#hooks-entry) – usually through `console_scripts` or
 `scripts` in setup.py.
 
+This language also supports `additional_dependencies`
+so it can be used with [local](#repository-local-hooks) hooks.
+The specified dependencies will be appended to the `pip install .` command.
```
asottile, line 389: probably remove the `.` from the command here

soxofaan: The `.` is in the actual command, so I'm just following that:
```python
    install_cmd = ('python', '-mpip', 'install', '.', *additional_dependencies)
```

https://github.com/pre-commit/pre-commit/blob/0252908c27b4e5539978ec8e676576a5e9ea8c2f/pre_commit/languages/python.py#L210

asottile: yes I know

soxofaan: To be honest, I'm not sure what you are suggesting.
I just wanted to update the docs to match the current implementation, but do I understand correctly that actually the implementation should be changed to something along the lines of the following (pseudo-code)?
```
if additional_dependencies:
    pip install additional_dependencies
else:
    pip install .
```

soxofaan: Ok I force-pushed it without the `.`

asottile (approved):

PR #985, opened 2024-08-01
DOC: clarify what "language" means in docs

tacaswell:
The use of "language" in the docs is miss leading, added some additional context and details.

I looked at a more significant edits, but decided to see if this minimal change would be accepted before doing more work.

### DIFF
--- a/sections/new-hooks.md
+++ b/sections/new-hooks.md
@@ -1,5 +1,7 @@
 pre-commit currently supports hooks written in
-[many languages](#supported-languages). As long as your git repo is an
+[many "languages"](#supported-languages).  In the context of pre-commit the term
+"language" refers to the package manager and environment isolation tools that pre-commit will use to
+install the hook.  As long as your git repo is an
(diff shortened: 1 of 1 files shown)


asottile:
I don't think this improves things but thanks for the patch anyway

tacaswell:
Do you have an alternative suggestion?   conda is not a language and virtualenv is not Python.

asottile:
I don't think anyone is actually confused about this.   I think you're being pedantic at best and wasting both of our time

tacaswell:
The information I was trying to find in the documentation was where the environments the hooks run live and how they are mode (which I did not find), but hit this page and was in fact confused for a bit.  What clarified things was the note on `python_venv` that implied that "python" meant "use virtualenv".

Even if you don't like the clarification at the top, I think adding the note to the python section to specify it uses virtualenv (out of the many possible ways to make isolated environments) is useful information to include.  I happy to modify this to only make that change if you will take it.

asottile:
I disagree.  it's an implementation detail at best and the answer to your actual question is already in the docs

PR #990, opened 2024-09-12
Add workaround script for pre-commit/pre-commit#1243

alexander-bauer:
I came across the issue described in pre-commit/pre-commit#1243 while using `podman-docker`, and developed this workaround. It seems like other pull requests trying to fix the issue have stalled, so I thought this would make sense as a note in the meantime.

Thank you all!


asottile:
hi thanks for the pr but I'm not going to include this hack here sorry

alexander-bauer:
I'm inclined to agree, I mainly wanted to leave a trail for anyone arriving through pre-commit/pre-commit#1243. I'll leave this workaround there, instead.

PR #1001, opened 2024-12-13
update pre-commit-hooks and black rev

Kirade:
- Updated pre-commit-hooks and black rev in install.md


asottile:
these aren't intended to be kept up to date (there are several similar rejected prs you can search for!) but are just examples

Kirade:
Oh I got it. Thanks!

PR #1003, opened 2024-12-31
docs: Update pre-commit-hooks version to v5.0.0

kang8:
When using pre-commit-hooks, I copied the version from the website.. But
when I actually use it, it reminds me:

```
[WARNING] repo `https://github.com/pre-commit/pre-commit-hooks` uses deprecated stage names (commit, push) which will be removed in a future version.  Hint: often `pre-commit autoupdate --repo https://github.com/pre-commit/pre-commit-hooks` will fix this.  if it does not -- consider reporting an issue to that repo.
```

So, the `pre-commit-hooks` version should be upgraded to the latest one.


asottile:
please look at other prs first next time!

PR #1007, opened 2025-01-29
Clarify pygrep behaviour when matching at the end of a string

nijel:
It is not obvious that newline is included in the matched line and it took me quite some time to figure out.

### DIFF
--- a/sections/new-hooks.md
+++ b/sections/new-hooks.md
@@ -489,6 +489,9 @@ For multiline matches, use `args: [--multiline]`.
 
 To require all files to match, use `args: [--negate]`.
 
+The single line matches include trailing newline, you need to include it to
+match something at the end of the string, for example `word\n$'.
+
 __Support:__ pygrep hooks are supported on all platforms which pre-commit runs
 on.
 


asottile:
this is already specified as a python regex so this behavior should be expected.  also `$` will match already without the new line being part of the pattern

nijel:
Hmm, it indeed does match for this simple case, but it doesn't work like that for all regexp. Sorry for providing a bad example.

What I wanted to catch:

```rst
* This is a correct changelog entry with a stop.
* This is an invalid entry
```

This doesn't work as expected because `[^.]` because is consumed by `\n`:

```
  - id: rst-bullet-stop
    name: Changes end with a stop
    description: Bullet list in changes should end with a stop.
    files: docs/changes\.rst
    entry: ^ *\* .*[^.]$
    language: pygrep
```

This does work:

```
  - id: rst-bullet-stop
    name: Changes end with a stop
    description: Bullet list in changes should end with a stop.
    files: docs/changes\.rst
    entry: ^ *\* .*[^.]\n$
    language: pygrep
```

asottile:
yeah that's just how regex works.  nothing special here

asottile:
your character class is too greedy

PR #1008, opened 2025-02-07
docs: add igorshubovych/markdownlint-cli to hooks page

davidxia:
The current Markdownlint hook is in Ruby.
Mention a popular Node-based one as well.

### DIFF
--- a/sections/hooks.md
+++ b/sections/hooks.md
@@ -75,13 +75,15 @@ for text / docs / prose:
 - [crate-ci/typos]: find and fix common typographical errors
 - [thlorenz/doctoc]: generate a table-of-contents in markdown files
 - [amperser/proselint]: A linter for prose.
-- [markdownlint/markdownlint]: a Markdown lint tool
+- [markdownlint/markdownlint]: a Markdown lint tool in Ruby
+- [DavidAnson/markdownlint-cli2]: a Markdown lint tool in Node
 - [codespell-project/codespell]: check code for common misspellings
 
 [crate-ci/typos]: https://github.com/crate-ci/typos
 [thlorenz/doctoc]: https://github.com/thlorenz/doctoc
 [amperser/proselint]: https://github.com/amperser/proselint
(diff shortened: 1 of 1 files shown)


asottile:
so I notice there's also https://github.com/DavidAnson/markdownlint-cli2 which appears to be written by the author of the underlying library

davidxia:
I also saw that. I’m not sure the differences between them. But I see Anson
is a regular contributor to the other one which has more than 500 stars
which is a requirement to be listed on the pre-commit hook page. Cli2 has
less than 500 stars.

asottile:
> I’m not sure the differences between them

I'd be more comfortable understanding the differences between choosing one for this page

davidxia:
A summary of the differences [is here](https://github.com/DavidAnson/markdownlint-cli2?tab=readme-ov-file#markdownlint-cli).

> * The glob implementation and handling of pattern matching is slightly different.
> * Configuration files are supported in every directory (vs. only one at the root).
> * The `INI` config format, `.markdownlintrc`, and `.markdownlintignore` are not supported.

A more detailed explanation is on [Anson's blog here](https://dlaa.me/blog/post/markdownlintcli2). He writes it's faster, config is more flexible, and fewer dependencies.

Based on this, I think listing Anson's is better. What do you think?

davidxia:
gentle bump

asottile:
I mean yeah seems fine

davidxia:
Thanks! Let me know next steps, if any, to get this merged.

asottile:
hmm I'm confused -- is your conclusion above that the `2` version is better?

davidxia:
> hmm I'm confused -- is your conclusion above that the `2` version is better?

Oh, sorry. Forgot to update the PR. Updated now

asottile:
> sorry, but just wanted to bump again. 🙏

don't do this.  I'll get to it when I get to it.  this isn't urgent

PR #1012, opened 2025-03-11
consolidate requirements and setup into pyproject toml - use uv for b…

clintonsteiner:
…uilds


asottile:
nobody asked for this

PR #1020, opened 2025-04-01
Expand on the phrase `no commits needed` for `try-repo` command

jaredkoontz:
I think this language should be improved. 

When I read it, I assumed this meant that you could just `git init` a repo and start developing a plug-in. It seems however, that pre-commit expects there to be a commit on the repo after discussions here:
https://github.com/pre-commit/pre-commit/pull/3441

### DIFF
--- a/sections/new-hooks.md
+++ b/sections/new-hooks.md
@@ -107,7 +107,7 @@ _note_: you may need to provide `--commit-msg-filename` when using this
 command with hook types `prepare-commit-msg` and `commit-msg`.
 
 Specifying a commit is not necessary to `try-repo` on a local
-directory, however, the repository must have at least one commit in its history. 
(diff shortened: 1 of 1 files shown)


asottile:
the original language is fine: in order for something to be tracked it has to have been committed

jaredkoontz:
as an end user, i read what is currently written: ```a commit is not necessary to `try-repo` on a local directory. ``` 

that leads me to believe when running `try-repo`, a commit is not necessary. One can interpret the phrase ```a commit is not necessary to `try-repo` on a local directory. ```  as create a git directory, throw in the .pre-commit-hooks.yml and start plugging away, which i did.

asottile:
sure if you stop reading there yes it could be misleading.  but that's why there's more words afterwards

PR #1039, opened 2025-09-29
Update install.md for pipx

ppbrown-bg:
Save non-python experts a lot of time, by mentioning that pipx exists, and is what ubuntu24 and later need instead of pip.

Otherwise, non-python programmers may spend a lot of time researching how to fix the problem, and worse yet, implement hacks and wrappers that make things ugly.


asottile:
please search for duplicates next time!

